PR #80, opened 2015-01-20
CAF::Object add report method to conditional logger

stdweird:


stdweird:
@jrha i set this milestone to 14.12 because this prevents the throwerror of https://jenkins1.ugent.be/job/NCM_Components-pull_requests/508/ (but the code shouldn't have entered that code path at all in the first place).

PR #84, opened 2015-03-26
Add TextRender usage documentation

stdweird:
This is the `CAF::TextRender` part of the tutorial given during the 19th workshop.

### DIFF
--- a/TextRender.md
+++ b/TextRender.md
@@ -263,6 +263,45 @@ Individual elements are dealt with via `CCM/pan_element.tt`
 [%- END -%]
 ```
 
-`CCM::TextRender` has the `doublequote` element option is set to produce
+The `doublequote` element option is set to produce
 a doublequoted string if `data` is a string and the `truefalse` option to
 generate `true` or `false` value if `data` is a boolean.
+
+So the following is possible:
+
+An object template
+
+```pan
+object template format;
+
+"/a" = 1;
+"/b" = 1.5;
+"/c/t" = true;
+"/c/f" = false;
+"/d" = "test";
+```
+
+with
+
+
+```perl
+my $trd = EDG::WP4::CCM::TextRender(
+    'CCM/pan',
+    $config->getElement('/'),
+    element => {
+        doublequote => 1,
+        truefalse => 1,
+    },
+);
+print "$trd";
+```
+
+gives
+
+```
+"/a" = 1; # long
+"/b" = 1.5; # double
+"/c/f" = false; # boolean
+"/c/t" = true; # boolean
+"/d" = "test"; # string
+```
--- a/TextRender.md
+++ b/TextRender.md
@@ -1,0 +1,162 @@
+# TextRender
+
+Generating structured text is best done with [`CAF::TextRender`][caf_textrender_docs].
+This document guides through the usage and testing of `CAF::TextRender`.
+
+Using `ncm-metaconfig`, which is the metacomponent build around
+`CAF::TextRender`, is described [here][metaconfig].
+
+TODO add correct/final url
+
+[caf_textrender_docs]: http://docs-test-caf.readthedocs.org/en/latest/CAF/CAF::TextRender
+[metaconfig]: https://github.com/quattor/configuration-modules-core/Metaconfig.md
+
+# CAF::TextRender
+
+Basic usage has 2 main modes:
+ * generate text : the `CAF::TextRender` instance has auto-stringification
+
+    <!-- language: lang-perl -->
+        use CAF::TextRender;
+        my $module = 'mymodule';
+        my $trd = CAF::TextRender->new($module, $contents, log => $self);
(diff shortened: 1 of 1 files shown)


TextRender.md
```diff
@@ -0,0 +1,162 @@
+# TextRender
+
+Generating structured text is best done with [`CAF::TextRender`][caf_textrender_docs].
+This document guides through the usage and testing of `CAF::TextRender`.
+
+Using `ncm-metaconfig`, which is the metacomponent build around
```
jrha: buil**t**

TextRender.md
```diff
@@ -0,0 +1,162 @@
...
+
+[caf_textrender_docs]: http://docs-test-caf.readthedocs.org/en/latest/CAF/CAF::TextRender
+[metaconfig]: https://github.com/quattor/configuration-modules-core/Metaconfig.md
+
+# CAF::TextRender
+
+Basic usage has 2 main modes:
+ * generate text : the `CAF::TextRender` instance has auto-stringification
+
+    <!-- language: lang-perl -->
```
jrha: Does this work? Shouldn't it use \`\`\`perl for consistency?

TextRender.md
```diff
@@ -0,0 +1,162 @@
...
+        use CAF::TextRender;
+        my $module = 'mymodule';
+        my $trd = CAF::TextRender->new($module, $contents, log => $self);
+        print "$trd"; # stringification
+
+ * write text to file : get a `CAF::FileWriter` instance with text from `CAF::TextRender` instance
+
+    <!-- language: lang-perl -->
```
jrha: Same question as above.

TextRender.md
```diff
@@ -0,0 +1,162 @@
...
+        use CAF::TextRender;
+        $module = "mymodule";
+        $trd = CAF::TextRender->new($module, $contents, log => $self);
+        my $fh = $trd->filewriter('/some/path');
+        die "Problem rendering the text" if (!defined($fh));
+        $fh->close();
+
+Besides the logger, the 2 main parameters are the `module` and the `contents`.
+The contents is a hash-reference with the data that is used to generate
```
jrha: \` around contents

TextRender.md
```diff
@@ -0,0 +1,162 @@
...
+```
+
+will generate
+
+```
+$ perl -e 'use Template; my $tttext="Hello [% world %]\n"; Template->new()->process(\$tttext, { world => "Quattor" });'
+Hello Quattor
+```
+
+TODO minimal version
```
jrha: Any idea what this is yet?

TextRender.md
```diff
@@ -0,0 +1,162 @@
...
+
+[TT_home]: http://www.template-toolkit.org/index.html
+
+TODO add url with examples
```
jrha: How about https://github.com/quattor/configuration-modules-core/tree/master/ncm-metaconfig/src/main/metaconfig as a start?

TextRender.md
```diff
@@ -0,0 +1,162 @@
...
+    );
+$rt->test();
+```
+
+With the default flags, each line is compiled as a multiline regular expression and matched against the text.
+The matches are also checked if they are ordered. In the example above `line 3` is expected to match in the text
+following `line 1`. But it does not need to be the next line (e.g. there could be a `line 2` in between).
+Each match is a test and each verification of the ordering also.
+
+TODO add correct/final url
```
jrha: ?

jouvin:
It would be good to have it finalized at the time of 15.4, even if not strictly part of the release...

TextRender.md
stdweird: That's for gihhub markdown, but generating documentation for readthedocs does understand github tags.

stdweird: i'll add the gh tags, we can figure out the readthedocs issue later

stdweird: i'm keeping the xml tags for now
see https://github.com/quattor/release/issues/86
this is each to search and replace later

TextRender.md
stdweird: the idea was to move all the readthe docs under one url, not one per repo. but i'll leave it as is

stdweird:
@jrha i'm adding a CCM::TextRender section too

stdweird:
@jrha i added the `CCM::TextRender` section, including an example that will only work if some of my branches get merged in 15.6 (in particular the `path` magic) (but are the best example of what `CCM::TextRender` can do)

TextRender.md
jrha: does understand github tags?
or 
doesn't understand github tags.

stdweird:
@jrha afaik, readthedocs does not understand gh tags (but i removed the tags to be consistent with the rest of the doc. i opened https://github.com/quattor/release/issues/86 to find a solution

jrha:
:crying_cat_face:

PR #85, opened 2015-04-06
TextRender: contents from Element instance and getTree convert options

stdweird:


stdweird:
requires https://github.com/quattor/CCM/pull/51

pom.xml
```diff
@@ -51,6 +51,9 @@
       <name>Ben Jones</name>
       <email>ben.jones@morganstanleylcom</email>
```
jrha: **.**com?

stdweird:
i'm having serious second thoughts about this PR. it's mixing CCM and CAF, and this breaks the dependency chain.
i'm probably going to move all CCM code to new subclass CCM::TextRender (as i would like to expose even more CCM details to TT usage; it would blow up completely).

stdweird:
There will be a CAF and a CCM PR to replace this

stdweird:
CCM part now in https://github.com/quattor/CCM/pull/51

PR #91, opened 2015-05-17
Application export cfgfile option name

stdweird:
Export the `cfgfile` option name as readonly `$OPTION_CFGFILE` (and use it internally in `CAF::Application`), mainly to prevent typos.

### DIFF
--- a/src/main/perl/Application.pm
+++ b/src/main/perl/Application.pm
@@ -320,6 +320,9 @@ sub _initialize ($$@) {
         }
     });
 
+    # initialise predefined options
+    $self->{'CONFIG'}->set($OPTION_CFGFILE, undef);
+
     # add application-specific options
     unless ($self->_add_options()) {
         throw_error('Cannot add options');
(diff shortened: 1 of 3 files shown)


jouvin:
I don't have a strong opinion on this one... As this PR is not associated with an issue, it would be good to put in the PR description the reason/use case for this change.

src/main/perl/Application.pm
```diff
@@ -302,160 +292,141 @@ Initialize the Application.
...
-  #
+    # check if we have a config file to read or a --help request.
+    # parse it 'by hand' instead of using AppConfig twice
+    # for cmd line parsing
+    my $help_request = 0;
+    my $configfile = undef;
+    my @args_tmp = @argv;
+    my $arg;
+
+    my $cfgfile_standalone_pattern = '^(--'.$OPTION_CFGFILE.')$';
```
ned21: I am not convinced by this change from hardcoding the word to using a constant and then a variable. It's definitely not more readable than before, possibly harder to follow than before.

ned21:
Unit tests and whitespace clean -> good. Not convinced by the addition of a constant for cfgfile.

jouvin:
@stdweird you say in the description "mainly to prevent typos"... Why it helps? (naive question)

stdweird:
yes, obviously this makes the `CAF::Application` code harder to read, but it is a constant and should always be the exact same value. but the main advantage is in the actual clients that need to set this option name. the constant is exported, and others should use it correctly.
e.g. https://github.com/quattor/CCM/blob/master/src/main/bin/ccm-fetch#L36 looks rather innocent, but all the magic comes from a specifc string `cfgfile`
we refer to them in our team as "magic strings".

@ned21 i'll see if i can make the regexps works without creating the intermediate patterns.

stdweird:
@ned21 hope it's more readable now

ned21:
Yep, code is more readable, thanks! 

I think the one-liner commit message could be clearer though. It's rather coded and terse - e.g. it's not clear that Application at the start refers to the file being changed rather than being an integral part of the sentence. I am OK with simple changes having brief commit messages but I think in this case a longer one might be called for.

stdweird:
@ned21 better now?

ned21:
Yep, thanks! Will let @jrha merge this when we start merging 15.6 items.

PR #98, opened 2015-08-07
FileWriter: disable chatty LC::Check by default unless NoAction

stdweird:
Get rid of pesky `updated /path/to/file` messages that are not logged anyway, and only pollute stdout.
It's still enabled when running with `--noaction`, since it might indicate what the application is modifying (and thus might be considered useful).
Requires https://github.com/quattor/ncm-ncd/pull/88 and https://github.com/quattor/ncm-cdispd/pull/42

### DIFF
--- a/src/main/perl/FileWriter.pm
+++ b/src/main/perl/FileWriter.pm
@@ -214,7 +214,15 @@ sub close
         *$self->{save} = 0;
         $str = *$self->{buf};
         *$self->{options}->{contents} = $$str;
-        $ret = LC::Check::file (*$self->{filename}, %{*$self->{options}});
+
+        my %cf_opts = %{*$self->{options}};
+        if(! exists($cf_opts{silent})) {
+            # make sure LC::Check::file is silent unless in noaction mode
+            # (or when explicitly set via silent option)
+            $cf_opts{silent} = *$self->{options}->{noaction} ? 0 : 1
+        }
+
+        $ret = LC::Check::file (*$self->{filename}, %cf_opts);
         # Restore the SELinux context in case of modifications.
         if ($ret) {
             $self->change_hook();
--- a/src/test/perl/test-caffileeditor.t
+++ b/src/test/perl/test-caffileeditor.t
@@ -68,6 +68,7 @@ is_deeply(\%opts, {
     mtime => 1234567,
     contents => TEXT,
     noaction => 0,
+    silent => 1,
 }, "options set in new(), derived noaction and current contents are passed to LC (via parent filewriter)");
 
 is(*$fh->{filename}, $filename, "The object stores its parent's attributes");


stdweird:
@ned21 please don't tell me MS has tools that rely on output being polluted with this?

src/main/perl/FileWriter.pm
```diff
@@ -181,7 +181,14 @@ sub close
         *$self->{save} = 0;
         $str = *$self->{buf};
         *$self->{options}->{contents} = $$str;
-        $ret = LC::Check::file (*$self->{filename}, %{*$self->{options}});
+
+        my %cf_opts = %{*$self->{options}};
+        # check file is silent unless in noaction mode
+        # (or explicitly set via silent option)
+        $cf_opts{silent} = *$self->{options}->{noaction} ? 0 : 1
```
ned21: A tertiary operator guarded by a post-hoc if() on a continuation line? Yuck! I think a full if() {} block would  be much more readable here.

ned21:
I don't recall seeing these messages except in legacy components calling LC::Check directly and not ones using CAF::FileWriter and I am not sure why given the code. In NoAction mode these are actually a bit redundant since with --verbose you get a diff, although without verbose it's still good to know what's being updated. Either way the message "I changed X/I would change X" needs to be logged. So now I am confused - do we have proper logging when CAF::FileWriter touches a file?

stdweird:
@ned21 when `FileWriter` touches a file, it does it via https://github.com/quattor/CAF/blob/master/src/main/perl/FileWriter.pm#L184. the `NoAction` bits are handled/supported by/passed on to `LC::Check`.
this PR only disables some `LC::Check` internal messaging via `print` in the `_message` method (https://github.com/quattor/LC/blob/master/src/main/perl/Check.pm#L130)
so `NoAction` will trigger the verbose logging in `FileWriter` based on the return value of `LC::Check::file`, the only differene is the logged message itself: `FileWriter` will always log 'was modified' or `was not modified`, while the `_message` method from `LC::Check` will change the message to `would modify` when `NoAction` is set.

stdweird:
retest this please

jrha:
@ned21 does that answer your question?

ned21:
I think so. What I was trying to remember is that it appears that with --verbose we display the diff and log that the file was updated (which also gets echo'd to screen). Without --verbose we display nothing. Do we log anything? It would appear [not because stdout is just an out of the logger, and verbose defaults to false](https://github.com/quattor/CAF/blob/master/src/main/perl/Reporter.pm#L59). This seems suboptimal to me: whether a file has been changed or not should always be logged, even if not shown on stdout. Making the diff optional seems OK, especially since we don't want to log sensitive information.

So this probably isn't related to this particular PR,  but I do feel what is logged should be improved and now you've opened the box....

stdweird:
@ned21 wrt the PR, do we keep the silent option false for NoAction (or do we make it silent altogether)?

wrt your other question: what is logged to file/syslog is the same as what is logged to stdout.
we can either change the logging to `info` for the checkfile result, or introduce a (new) log-related option for which i have no name yet that allows for a different loglevel for syslog/logtofile, with the clarification that this new loglevel is always the max of the general loglevel and whatever the new one is set to, to avoid missed log information. this way you can log to syslog with `verbose` and keep stdout civil (e.g. on `info`).
but can you open an issue for any further discussion on that? :smile:

ned21:
I think we keep it in NoAction mode otherwise --noaction mode without --verbose is pretty useless. (At least until #100 is fixed).

If there's agreement that we should just log the checkfile result at info level, is it worth merging this as a temporary workaround, or just make the proper fix?

stdweird:
if `--noaction` without `--verbose` is pretty useless, we can change the default loglevel based on NoAction being set (well, i could have a look :smile:)
the whole point of the PR was to reduce what is printed to stdout, e.g. in the `ccm` tool. we could change the tools to start at loglevel warn ofcourse. as the modification messages were printed and not logged, i assumed noone was really interested that much. for `--noaction` they certainly make sense, hence the code in the PR.

jrha:
As discussed at standup today, @stdweird is working on a "proper" fix for #100.

jrha:
@stdweird rebase?

jrha:
@stdweird *poke* as well

stdweird:
@jrha i've moved the milestone for now. and i'm not rebasing until some other issues have been implemented. we're debugging a permissions related problem, that was spotted using the default chatty behaviour.

jrha:
Okay, sounds good :snail:

piojo:
retest this please

ned21:
As discussed, do not merge until other changes are merged so this can be tested with them.

stdweird:
@ned21 what needs to be tested? `verbose_logfile` with and without noaction works as intended, but https://github.com/quattor/ncm-ncd/pull/88 needs merging before this can be merged

ned21:
@stdweird Sorry I must have mis-heard/mis-understood/mis-transcribed what you said to me. Whatever you think needs to be done.

stdweird:
if https://github.com/quattor/ncm-ncd/pull/88 gets merged in 16.10, the logfiles should contain all info required. i've put the milestone on the the next release in case we missed something.

PR #99, opened 2015-08-12
FileEditor:do not insert the missing newline if the to-be-inserted text starts with a newline

stdweird:


stdweird:
@jrha needed for spma

Issue #100, opened 2015-08-14
CAF::FileWriter does not log file changes


ned21:
Follow up from discussion in #98. 

With --verbose we display the diff and log that the file was updated (which also gets echo'd to screen). Without --verbose we display nothing. Do we log anything? It would appear not because stdout is just an out of the logger, and verbose defaults to false. This seems suboptimal to me: whether a file has been changed or not should always be logged, even if not shown on stdout. Making the diff optional seems OK, especially since we don't want to log sensitive information.

A simple fix would be to log the result of the Check::File at info level. Given that until #98 this was being printed to stdout it would not be a net increase in the amount of lines shown on stdout during an interactive run.

stdweird:
@ned21 i'd prefer not raise the default to `info`. i'll have a look to split up the loglevel for logtofile/syslog and stdout (or something like that).

ned21:
Logging verbose message to the file by default but not to stdout is fine with me. We always try to be generous with what we log to a file and conservative with what we log to stdout.

stdweird:
@ned21 @jrha i've changed my mind a bit how to handle this. i still tend to implement the splitted loglevels, but i'm afraid im'm going to reinvent Log4Perl or something like that. 
would it be nicer if https://github.com/quattor/ncm-ncd/issues/14 worked? well, i tried to get some part resurrected by https://github.com/quattor/CAF/pull/117. I'll try to implement the relevant part in ncm-ncd such that ncm-ncd generates an overview of the modified files. would that suffice?

stdweird:
@ned21 this is now stdout with #117, #98 and https://github.com/quattor/ncm-ncd/pull/51
```
2015/10/12-12:09:53 -----------------------------------------------------------
2015/10/12-12:09:54 [INFO] configure on component metaconfig executed, 0 errors, 0 warnings
2015/10/12-12:09:54 [INFO] EVENT: metaconfig modified file /etc/accountpage_db.conf
```

so no more `updated` polution, and each component logs a nice line (ofcourse, that specific format is just a proposal)

piojo:
> but i'm afraid im'm going to reinvent Log4Perl or something like that.

So, why not replace CAF::Log with Log::Log4perl (or make CAF::Log use
Log::Log4perl instead of LC::Log)?

stdweird:
@piojo instead of `LC::Log`? it's not like it uses it atm :smile: 
but :+1: on replacing it with Log4Perl

ned21:
[...]

Issue #93, opened 2015-07-07
Add CAF::Log option to record the PID


msmark:
Look at the following snippet from the `ncd.log` file:

```
2015/07/06-14:06:26 [INFO] running component: nss
2015/07/06-14:06:26 ---------------------------------------------------------
2015/07/06-14:06:35
2015/07/06-14:06:35 ------------------------------------------------------------
2015/07/06-14:06:35 [INFO] NCM-NCD version 1.4.1 started by root at: Mon Jul  6 14:06:35 2015
2015/07/06-14:06:39 [WARN] signal handler: received signal: INT
2015/07/06-14:06:39 [ERROR] ncd exiting gracefully after signal hit.
2015/07/06-14:09:52
2015/07/06-14:09:52 =========================================================
2015/07/06-14:09:52
2015/07/06-14:09:52 [OK]   0 errors, 0 warnings executing configure
```

With a little bit of guesswork and intuition, you can figure out what’s going on.  But it would be easier to read if the PID of the `ncm-ncd` process was included, like this:

```
[1273] 2015/07/06-14:06:26 [INFO] running component: nss
[1273] 2015/07/06-14:06:26 ---------------------------------------------------------
[1280] 2015/07/06-14:06:35
[1280] 2015/07/06-14:06:35 ------------------------------------------------------------
[1280] 2015/07/06-14:06:35 [INFO] NCM-NCD version 1.4.1 started by root at: Mon Jul  6 14:06:35 2015
[1280] 2015/07/06-14:06:39 [WARN] signal handler: received signal: INT
[1280] 2015/07/06-14:06:39 [ERROR] ncd exiting gracefully after signal hit.
[1273] 2015/07/06-14:09:52
[1273] 2015/07/06-14:09:52 =========================================================
[1273] 2015/07/06-14:09:52
[1273] 2015/07/06-14:09:52 [OK]   0 errors, 0 warnings executing configure
```

This example occurs when the user manually launches `ncm-ncd` shortly after the dispatcher decides there is work to do.

Can we add an option to `CAF::Log` so the PID can be recorded, and then modify `ncm-ncd` to use that option when setting up the logger for `ncd.log`?

PR #103, opened 2015-09-03
Cleanup and unittest Object, Log, Reporter and ReporterMany

### LINKED ISSUE #102: remove ReporterMany (or explain what is supposed to add to Reporter)
code is only used in `ncm-ncd`, but i fail to see what it should do differently than `Reporter`.

stdweird:
This is preparation for #100 and #93
Fixes #102 (not via removal, but through proper subclass of `Reporter`)

minor backwards incompatibility:
* subclassed ReporterMany has different/fixed/correct `setup_reporter` behaviour
 - be able to turn off quiet and verbose                                                                                                            
 - verbose triggered by debug                                                                                                                       
 - init with global _REP_SETUP happens on usage via _rep_setup method, not on `setup_reporter` (i.e. it will report things without `setup_reporter` being called if global `Reporter::_REP_SETUP` is set to do so).


stdweird:
@ned21 any toughts on this? (esp the modified log format)

src/main/perl/Log.pm
```diff
@@ -26,75 +41,70 @@ CAF::Log - Simple class for handling log files
...
-
 =head2 Public methods
 
 =over 4
 
-=item close(): boolean
+=item C<close()>: boolean
 
-closes the log file.
+closes the log file, returns SUCCESS on succes, undef otherwise
```
ned21: typo: success

src/main/perl/Log.pm
```diff
@@ -26,75 +41,70 @@ CAF::Log - Simple class for handling log files
...
+(if no FH attribute exists).
 
 =cut
 
-sub close ($) {
-  my $self=shift;
+# Called during DESTROY, use $_XYZ flavour
+sub close ($)
```
ned21: Previous commits have removed prototypes from function - do the same here?

src/main/perl/Log.pm
```diff
@@ -26,75 +41,70 @@ CAF::Log - Simple class for handling log files
...
+{
+    my $self = shift;
 
-  return undef unless (defined $self->{'FH'});
-# Why adding extra newlines???
-#   $self->{'FH'}->print("\n");
-  $self->{'FH'}->close();
-  $self->{'FH'} = undef;
+    return unless (defined $self->{$_FH});
```
ned21: Is it useful to indirect the attribute names in this way? What's wrong with just ```$self->{FH}``` ?

ned21:
@stdweird sorry it took my so long to get round to reviewing this it seems the new log format was reverted anyway?

src/main/perl/Log.pm
stdweird: i'd prefer to do that outside of thei work, i opened https://github.com/quattor/CAF/issues/114

src/main/perl/Log.pm
stdweird: its mainly to prevent typos. if this has a typo, perl will not complain.
also, `FH` in your example could be the string `FH` or a constant.
 
the `$_FH` is special due to DESTROY quirks i ran into while testeing this.

stdweird:
@ned21 thanks for reviewing
wrt the log format, i opened #115. i'm fine with keeping the old format, just wanted to point out this (small) inconsistency

stdweird:
@ned21 @jrha anything left to do?

jrha:
LGTM

stdweird:
@ned21 ping

src/main/perl/FileWriter.pm
```diff
@@ -173,8 +172,8 @@ sub close
                                   keeps_state => 1);
         $cmd->execute();
         *$self->{LOG}->verbose ("Changes to ", *$self->{filename}, ":");
+        $diff = "" if (! defined($diff));
```
ned21: Does this cause report to log an empty line or is it suppressed by the logging stack? Is an empty line the right way of indicating an empty diff or should we print something standardised such as "(no changes)" ?

ned21:
Aside from the one question I think this is good to merge.

src/main/perl/FileWriter.pm
stdweird: this change is to get rid of some 'undefined' warnings (and i think only in the unittests). there's also https://github.com/quattor/CAF/pull/119 to handle empty diffs

PR #106, opened 2015-09-10
Fix failing unittests on EL5 due to missing mktemp --tmpdir option

stdweird:


stdweird:
@jrha detected by build_all_repos

PR #110, opened 2015-09-19
Download: API proposal

stdweird:
There is no funcitional code here, only an API.


stdweird:
@ned21 @jrha this is a PR to define the interface to use and have it reviewed/discussed before adding more work.

stdweird:
retest this please

src/main/pan/quattor/types/download.pan
```diff
@@ -0,0 +1,131 @@
+declaration template quattor/types/download;
+
+include 'pan/types';
+
+@documentation{
+    A string that represents a URL that can be handled by CAF::Download
+    Format: [auth+][method+]protocol://location
+    Protocols: http|file
```
ned21: https as well?

src/main/pan/quattor/types/download.pan
```diff
@@ -0,0 +1,131 @@
...
+    Methods: lwp|curl transport
+    Auth: kinit|gssapi|x509 authentication
+    Location: anything
+}
+type caf_url_string = string with {
+    mp_l = split('://', SELF);
+
+    if (length(mp_l) != 2) {
+        error('invalid URL string requires ://, got' + SELF);
+        return(false);
```
ned21: redundant - error() already exits

src/main/pan/quattor/types/download.pan
```diff
@@ -0,0 +1,131 @@
...
+    };
+
+    if(! match(mp_l[0], '^((kinit|gssapi|x509)\+)?((lwp|curl)\+)?(https?|file)$')) {
+        error('invalid method+protocol for ' + mp_l[0]);
+        return(false);
```
ned21: ditto

src/main/perl/ObjectText.pm
```diff
@@ -0,0 +1,309 @@
...
+    }
+
+And use it via
+    my $sc = SubClass->new(log => $self);
+    print "$sc"; # stringification
+
+    $sc = SubClass->new(log => $self);
+    # return CAF::FileWriter instance (text already added)
+    my $fh = $sc->filewriter('/some/path');
+    die "Problem rendering the text" if (!defined($fh));
```
ned21: Is this proper QWG programming style to die() like this? I thought we just called $self->error() and continued?

src/main/perl/Download/Retrieve.pm
```diff
@@ -0,0 +1,153 @@
...
+        my $url = shift @urls;
+
+        my $txt = $url->{_string};
+        my $id = $url->{_id};
+
+        $self->verbose("download url $txt (id $id) attempt ",
+                       ($tried{$id} || 0) + 1,
+                       " total attempts $tries.");
+
+        # TODO: should we really wait if we try another url?
```
ned21: For a download that has N urls, they should be tried in sequence without delay, and only after all N have failed should there be any back off & wait. [IMHO]

src/main/perl/Download.pm
```diff
@@ -0,0 +1,142 @@
...
+
+# TODO: dependencies on curl and kinit
+
+Readonly::Hash my %DOWNLOAD_METHODS => {
+    http => [qw(lwp curl)], # try https, if not, try http
+    https => [qw(lwp curl)], # only https
+    file => [qw(lwp)],
+};
+
+Readonly::Array my @DOWNLOAD_PROTOCOLS => sort keys %DOWNLOAD_METHODS;
```
ned21: Why is that sorted?

src/main/perl/Download.pm
```diff
@@ -0,0 +1,142 @@
...
+
+use Readonly;
+use LC::Exception qw (SUCCESS);
+
+our @EXPORT_OK = qw(set_url_defaults);
+
+# TODO: dependencies on curl and kinit
+
+Readonly::Hash my %DOWNLOAD_METHODS => {
+    http => [qw(lwp curl)], # try https, if not, try http
```
ned21: Sorry, I don't follow the comment, ```# try https, if not, try http``` ?

src/main/perl/Download.pm
```diff
@@ -0,0 +1,142 @@
...
+=head1 NAME
+
+CAF::Download - Class for downloading content from remote servers.
+
+=head1 SYNOPSIS
+
+    use CAF::Download;
+
+    my $dl = CAF::Download->new(['https://somewhere/myfile']);
+    print "$dl"; # stringification
```
ned21: Some detail on how the user/caller should handle errors would be helpful.  e.g. how do I get a meaningful string to put in my call to $self->error() if I am supposed to do that myself? Don't other CAF::* methods do that on my behalf?

(Sorry if I am a bit confused: the underlying classes look OK but I don't quite yet see a connection from here to them.)

src/main/pan/quattor/types/download.pan
stdweird: wel, yes and no. i'm confused myself about the naming with all the available combinations and tools.

there are a number of "scenario's" i want to support, but not sure splitting them up in protocol/method/auth makes sense
* file: read from file (could require authentication like `kinit+file`, e.g. reading from afs (although i'm not sure if that isn't authorisation instead of  authentication))
* http: non-encrypted read from server using `HTTP` `GET` etc etc (could require authentication, like 'Basic auth', but is pretty non-safe, so there's no authentication  called `passwd` or something). 
 * plain http like MS uses this, i.e. by encrypting the payload, it becomes secure (authentication via decryption and encryption), and it requires a decrypting postprocessing otherwise the downloaded data is garbage. so calling this protocol http is maybe also not ok. i was thinking to put this under `kinit+http` (or `gssapi+http`) and detect the `MS` scenario via the same trick that CCM uses (the `krbencrypt` encoding https://github.com/quattor/CCM/blob/master/src/main/perl/Fetch/Download.pm#L96; i guess `krbencrypt` is something MS made up or is it a standardised name?)
 * https implies a valid client certificate, so `x509+http` is basically https (https without a valid client certificate is just not smart)
 * and there's the stuff i want to add for freeipa: http + authentication via spnego and kerberos, which is another flavour of `kinit+http` or `gssapi+http`, so a bit confusing with MS scenario, so i might call this `kinit+httpn` (`HTTP Negotiate` is a term used by microsoft for this, but they also support ntlm as authentication).
* in the future, i would also add `kinit+ipavault` or other custom tools used to manage sensitive data

the first 3 are currently supported by `ccm-fetch` and `ncm-download` (afai understand the code). the 4th one is to address https://github.com/quattor/CCM/issues/54, the 5th one is there so we don't have to try to setup our own secure datastore and be able to use existing tools.

the difference between `kinit` and `gssapi` mainly lies in the way the token is generated/managed: `kinit/kdestroy` binary or perl `GSSAPI`, but they both mean 'use kerberos`.

src/main/perl/Download.pm
stdweird: for a brief momenet i tought about allowing the quivalent of `wget --no-check-certificate`, i know it's not http, but i also don't consider it https (altough it is encrypted)

src/main/perl/Download.pm
stdweird: because it's an array, and for reproducability it should be sorted. but the sorting has no meaning, and it seems the current code doesn't use this anymore.

src/main/perl/Download.pm
stdweird: i'll address error handling in the `ObjectText` class (but i have to admit this really is not trivial in `CAF`, i think i've only implemented this once myself).
it would actually be nice to know what is expected. is it ok for a `CAF` class to throw exceptions? most of current code do not handle exceptions being thrown

src/main/perl/ObjectText.pm
stdweird: well, there's a huge difference between `$self->error()` and `die`: `die` stops, `error` just continues. i think the proper QWG would be 
```perl
if (!defined($fh)) {
    $self->error("Problem with text $sc->{fail}");
    return;
}
```
but that doesn't fit on a single line :smile:

src/main/perl/Download.pm
ned21: It's not obvious how that is related to lwp & curl. Falling back to an insecure method by default is bad but having it configurable is good.

src/main/perl/Download.pm
ned21: No idea, sorry. I have never taken a close look at the LC exception stuff and how it should be used.

stdweird: can we move this discussion to #120 ?

jrha:
@stdweird where did this get to?

jrha:
Still a good idea, several components still implement their own download mechanisms.

Issue #108, opened 2015-09-11
get rid of TextRender format general


stdweird:
The `Config::General` module should not be used to render configfiles, as it's can't guarantee consistent and correct output.

stdweird:
see #107 and earlier #97

stdweird:
@ned21 @piojo seems to have some recollection you are intersted or using this?

ned21:
We have a couple of places where we are using it, mainly to create files in /etc/modprobe.d. I need to look into whether we can swap those out with a modern version of ncm-modprobe or not, although ISTR you wanted to move away from things like ncm-modprobe and replace it with ncm-metaconfig, in which case, what's the recommended way forward?

(I sent some example configs to Jindrich to share with you f2f.)

stdweird:
@ned21 we should create a small TT file that produces the format. i'll check with jindrich how hard that can be. are you using `ncm-metaconfig` for it right now?

ned21:
We have a bit of a haphazard collection of different approaches atm but we want to standardise on the "right" way to manage these files.

stdweird:
`ncm-modprobe` is still needed for the `mkinitrd`, but should be expanded to support multiple files, and maybe even proper schema. however this is beyond this PR imho.

https://github.com/quattor/CCM/pull/73 should provide the `<key> <value>` you need for the examples mentioned above

PR #116, opened 2015-10-05
TextRender: remove builtin general (provided via CCM::TextRender)

stdweird:
Related to #108, functionality provided by https://github.com/quattor/CCM/pull/73

### DIFF
--- a/src/main/perl/TextRender.pm
+++ b/src/main/perl/TextRender.pm
@@ -21,7 +21,6 @@ use JSON::XS;
 use YAML::XS;
 use Config::Properties;
 use Config::Tiny;
-use Config::General;
 
 use base qw(CAF::Object Exporter);
 
@@ -104,7 +103,7 @@ CAF::TextRender - Class for rendering structured text
     my $trd = CAF::TextRender->new($module, $contents, log => $self);
     print "$trd"; # stringification
 
-    $module = "general";
+    $module = "yaml";
     $trd = CAF::TextRender->new($module, $contents, log => $self);
     # return CAF::FileWriter instance (rendered text already added)
     my $fh = $trd->filewriter('/some/path');
@@ -154,15 +153,11 @@ Java properties format (using C<Config::Properties>),
 
 .INI format (using C<Config::Tiny>)
 
-=item general
-
-(using C<Config::General>).
-
-Caution: there is no way to garantee reproducible results as
-there is no control over the order of hash keys. It is recommended to
-try to generate this format in another way (e.g. via TT).
-
 =back
+
+(Previously available module <general> was removed in 15.12.
+Component writers needing this functionality can use
+the L<CCM::TextRender> subclass instead).
 
 Or, for any other value, C<Template::Toolkit> is used, and the C<module> then indicates
 the relative path of the template to use.
@@ -179,7 +174,7 @@ It takes some extra optional arguments:
 
 =item C<log>
 
-A C<CAF::Reporter> object to log to.
+A L<CAF::Reporter> object to log to.
 
 =item C<includepath>
 
@@ -266,8 +261,8 @@ sub _initialize
     # set render method
     ($self->{method}, $self->{method_is_tt}) = $self->select_module_method();
 
-    # set contents, after module is selected (some modules trigger
-    # allow module aware contents changes
+    # set contents, after module is selected (some modules
+    # allow module aware contents changes)
     $self->{contents} = $self->make_contents();
 
     return SUCCESS;
@@ -662,15 +657,6 @@ sub render_tiny
         }
     }
     return $c->write_string();
-}
-
-
-sub render_general
-{
-    my ($self) = @_;
-
-    my $c = Config::General->new(-SaveSorted => 0);
-    return $c->save_string($self->{contents});
 }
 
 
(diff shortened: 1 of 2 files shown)


src/main/perl/TextRender.pm
```diff
@@ -154,16 +153,10 @@ Java properties format (using C<Config::Properties>),
...
-
-(using C<Config::General>).
-
-Caution: there is no way to garantee reproducible results as
-there is no control over the order of hash keys. It is recommended to
-try to generate this format in another way (e.g. via TT).
-
 =back
 
+(Builtin module <general> is provided as part of L<CCM::TextRender> since 15.10).
```
ned21: I don't follow this comment, sorry. So "general" is still a valid module? How is CCM:TextRender invoked since that depends on this class, not the other way? 

<guessing>Is there some magic going on related to the use of any other value passed to module causes it to look for a TT file of that name, and we always provide a general.tt file? If so, I think it would be clearer to remove this parenthesised "afterthought" sentence and rewrite line 160 to explain how general is a special case of the TT functionality.

stdweird: well, `general` is now an alias for `CCM/general.tt` (https://github.com/quattor/CCM/pull/73/files#diff-3a3731fb118515c51d43b8562ff39e47R316), so if you're using CCM::TextRender (like metaconfig and the test framework), nothing should have changed.

the pod of CCM::TextRender now explains what general is. Maybe i should drop the word `buitin`?

ned21: If general is still available  as a "module" and useable in the same was from a user perspective then it should remain an =item, and yes drop "builtin". Something like:

```
=item general

key = value\n format, provided by L<CCM::TextRender>. See the CCM::TextRender pod for more information.
```

jrha:
@stdweird?

src/main/perl/TextRender.pm
stdweird: `general` is not available in `CAF::TextRender`, only in `CCM::TextRender`, which is a subclass of `CAF::TextRender`.  i don't mind readding the item, but i think this is confusing as e.g. following code will not work
```perl
use CAF::TextRender;

my $trd = CAF::TextRender->new('general', {some => 'hashref'});
# do something useful with $trd
```

jrha:
This looks fine to me now...

src/main/perl/TextRender.pm
ned21: OK, so I would just note that the previously available general module was removed in release X. If CCM::TextRender and/or a module called "general" isn't generally available then it's confusing to mention it in any context other than "this no longer works"

ned21:
I disagree. :-) The POD is still confusing and ambiguous...

src/main/perl/TextRender.pm
stdweird: something like
```
(Previously available module <general> was removed in 15.12, and is now provided as part of L<CCM::TextRender>).
```
Or do you prefer this in an `=item general`?

ned21: Is the old functionality available via the CCM::TextRender module? While CAF::* is a public API, I would have thought CCM::TextRender is a private API for manipulating CCM files so we shouldn't imply that it's useable. In which case I would just stop at: "... was removed in 15.12, see http://github.com/issue/... for details."

Or is it intended that component writers can call upon the functionality available in CCM::TextRender if they wish? In which case I would say: "... was removed in 15.12. Component writers needing this functionality can use the L<CCM::TextRender> subclass instead." (But that seems strange - why did we ditch it if it's still available?)

stdweird: define "old fucntionality". the old functionality relies on an external perl library `Config::General`, but the output is not reproducable (no sorting of keys and all that) and has some strange quirks too (https://github.com/quattor/CAF/issues/108 and refereneces therein). this leads to not being able to run unittests reliably, which is a problem.
so the most common part of the `Config::General` format is now produced using TT in (https://github.com/quattor/CCM/pull/73/files#diff-d8db3099576a834ae6c4de684c93ec08R1 in PR https://github.com/quattor/CCM/pull/73) and is called `general`. However, the TT code relies on some features that are made available via the internals of `CCM::TextRender` (not explaining this here, we can try to setup a confcall if needed. Strictly speaking, the parts required to make the TT work, could be ported back to CAF, but at the expense of people being confused and not knowing how to use it). My overall recommendation is not to use `CAF::TextRender` directly in any quattor code at all, use `CCM::TextRender` instead (https://github.com/quattor/quattor.github.com/blob/master/_posts/2015-09-20-textrender-basics.md#textrender-caf-and-ccm, esp the last remark of that paragraph). ncm-metaconfig and a whole bunch of other components already do this. 

So `CAF::TextRender` does not provide the `general` module anymore. If you want to use it, you can, but you have to use `CCM::TextRender`. So if you have a component that uses the general format, you have to do `s/CAF::TextRender/EDG::WP4::CCM::TextRender/g', for usage of general via metaconfig, nothing is required; this is assuming https://github.com/quattor/CCM/pull/73 also gets merged in the same release.

I'll adapt the formulation if you agree:
```perl
(Previously available module <general> was removed in 15.12. Component writers needing this functionality can use the L<CCM::TextRender> subclass instead).
```
(I'm afraid the conversation will get lost once i push a commit that modifies this line again.)

jrha:
Wasn't it always?

src/main/perl/TextRender.pm
ned21: Thanks for the detailed explanation. This area is much more complex than I had remembered - the link to the blog post was a handy refresher. I agree with the formulation you propose.

stdweird:
@ned21 reformulated (and conversation lost :unamused: )

PR #117, opened 2015-10-10
History: support event tracking

stdweird:
Based on #120

### DIFF
--- a/src/main/perl/History.pm
+++ b/src/main/perl/History.pm
@@ -209,6 +209,50 @@ sub event
 
 =pod
 
+=item query_raw
+
+Primitive interface to query the events.
+
+C<match> is a anonymous sub that is passed
+the event as (only) argument
+(each event is a metadata hashref).
+Returns true if the event matches and is to be returned.
+
+C<filter> is an arrayref of metadata keys to filter from the event
+(only event metadata matching the filter is returned).
+
+Returns an arrayref of (a shallow copy of) the event metadata.
+
+TODO: support proper, human-friendly query interface via (NO)SQL
+
+=cut
+
+sub query_raw
+{
+    my ($self, $match, $filter) = @_;
+
+    my @res;
+    foreach my $ev (@{$self->{$HISTORY}->{$EVENTS}}) {
+        if ($match->($ev)) {
+            my $res_ev;
+            if($filter) {
+                # only add existing attributes, otherwise a hashslice would be better
+                my %f_ev = map { $_ => $ev->{$_} } grep { exists $ev->{$_} } @$filter;
+                $res_ev = \%f_ev;
+            } else {
+                # shallow copy
+                $res_ev = { %$ev };
+            }
+            push(@res, $res_ev);
+        }
+    }
+
+    return \@res;
+}
+
+
+=pod
+
 =item close
 
 Closes the history which triggers following
--- a/src/main/perl/History.pm
+++ b/src/main/perl/History.pm
@@ -12,7 +12,7 @@ use LC::Exception qw (SUCCESS);
 use Readonly;
 
 use parent qw(Exporter);
-our @EXPORT_OK = qw($IDX $ID $TS $REF);
+our @EXPORT_OK = qw($EVENTS $IDX $ID $TS $REF);
 
 # refaddr was added between 5.8.0 and 5.8.8
 use Scalar::Util qw(blessed refaddr);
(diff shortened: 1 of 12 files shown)


src/main/perl/History.pm
```diff
@@ -0,0 +1,394 @@
...
+use strict;
+use warnings;
+
+use LC::Exception qw (SUCCESS);
+use Readonly;
+
+use parent qw(Exporter);
+our @EXPORT_OK = qw($EVENTS $IDX $ID $TS $REF);
+
+# refaddr was added between 5.8.0 and 5.8.8
```
ned21: Should this line have a "use 5.8.8" line in it to help catch versions of Perl that ar too old?

src/main/perl/Log.pm
```diff
@@ -106,82 +116,94 @@ sub print ($$) {
...
+=item C<_initialize($filename, $options)>
+
+C<$options> is a string with magic letters
+
+=over
+
+=item a: append to a logfile
 
-initialize the object. Called by new($filename,$options).
+=item w: truncate a loglfile
```
ned21: loglfile => logfile

jrha:
@stdweird?

stdweird:
@jrha i fixed the remarks, but i'd prefer #103 to be merged first. than @ned21 can have another look :smile:

stdweird:
@ned21 rebased for easier review

ned21:
I think the code is good but it's hard to see how this actually useful without a working POC application. My concern is that a lot of CAF/LC is already really (too?) complicated that this adds more code where things can go wrong unexpectedly. 

Can we try sketching up how this might be used to implement some sort of "delete if removed from profile" option in ncm-metaconfig?

stdweird:
What exists now in https://github.com/quattor/ncm-ncd/pull/51 is that at the end of each component run, it prints all events that came from a FileWriter (and Editor i think) and that modified something.
but it has the events of all files handled by FileWriter, including the ones that didn't change, so we can track what is possibly modified by metaconfig (or any other component).

What is is keeping some sort of local database to keep track of which files where once handled by a FileWriter (or a FileWriter/component combination). at the end of each component, the list of files once handled by FileWriter can be compared to list of files handled by FileWriter, and any differences can be acted upon.

So one missing bit is to store the recorded events at the end of each component (or ncm-ncd), and provide a way to query the stored events.
The other missing bit is to define actions on certain types of events: e.g. do we remove a file that is under metaconfig control, but that was not modified ever? e.g. configfile from rpm, and metaconfig just creates the same config file. if there's a backup, we could restore the backup instead of just deleting, but is the backup the backup of the pre-metaconfig run or just the latest backup.

same applies for e.g. symlinks: need to create a CAF::Symlink module, and add an event handler to the init and/or the createing of symlinks, and then we can act upon these symlink events.

(none of database or actions to take is implemneted here because this PR is mainly to handle https://github.com/quattor/CAF/pull/98)

stdweird:
Had to rebase this on #120, there was a conflict

src/main/perl/History.pm
jouvin: We could but as we don't support anything older than EL5, it should always match...

ned21:
Thanks for the context -- I am not sure I had realised the significance wrt to #98 before. Can we defer this one until the next workshop? I'd like to see a working demo of quattor/ncm-ncd#51 and have a full session on what we want and expect from logging both at a library and component level to avoid a few of us going off in a direction the rest of the community disagree with.

stdweird:
@ned21 anything in particular you want to see in the demo? i could try to arrange access to one of our systems, we are running this already.
but i agree we should discuss (and hopefully agree) on the way forward at the next workshop.

ned21:
@stdweird - an example of what works right now would be good, along with mock ups of anything you envision for the future. That should provide some good fuel for either violent agreement or polite debate. :-)

jrha:
If this needs to be discussed at the next workshop, it won't be ready for 16.2?

brunocanning:
As an aside, would it be worth adding some functionality whereby the component first compares the file on the host with the profile that is one number less than the current profile number before making changes to the file? If a difference is found, it could log a warning that the file may have been modified by something other than quattor.

We occasionally have a problem at RAL were a person modifies by hand a file that is under quattor control and gets upset when their changes get overwritten by quattor. Whilst this behaviour is against our policy, it does happen and it would be nice to have a stick to use against these people.

stdweird:
@brunocanning can you open an issue for that feature?

jrha:
@stdweird rebase?

jrha:
@brunocanning please reference this PR when you open the issue.

jrha:
@stdweird *poke*

stdweird:
@jrha rebased. this is safe to merge, because it doesn't do much on its own (as ooposed to the chatty LC pr)

jrha:
Weird, Github is still telling me this is out of date, but that I can merge anyway... have they changed something again?

stdweird:
@jrha :rebased_on_master:

jrha:
:grinning:

PR #118, opened 2015-10-14
Service: support subclassing

stdweird:
Make `CAF::Service` a possible parent class for component specific services.
(E.g. https://github.com/quattor/configuration-modules-core/pull/620)

Fixes quattor/configuration-modules-core#608

### DIFF
--- a/src/main/perl/Service.pm
+++ b/src/main/perl/Service.pm
@@ -11,9 +11,10 @@ use CAF::Process;
 use LC::Exception qw (SUCCESS);
 
 our $AUTOLOAD;
-use base qw(CAF::Object);
+use base qw(CAF::Object Exporter);
 
 use constant DEFAULT_SLEEP => 5;
+use constant FLAVOURS => qw(linux_sysv linux_systemd solaris);
 
 # Mapping the methods we expose here to the svcadm operations. We
 # choose the Linux terms for our API.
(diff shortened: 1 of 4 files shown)


stdweird:
@piojo you might want to have a look at this. i especially like how the regular service unittests do not tets autoloading becasue they depend on what the set_service_variant does.
i tackled that in https://github.com/quattor/maven-tools/pull/66 (and called more appropriate force_service_variant)
the service-subclass tests however don't have any magic and (afaik) properly test the autoloading

src/main/perl/Service.pm
```diff
@@ -11,9 +11,10 @@ use CAF::Process;
 use LC::Exception qw (SUCCESS);
 
 our $AUTOLOAD;
-use base qw(CAF::Object);
+use base qw(CAF::Object Exporter);
 
 use constant DEFAULT_SLEEP => 5;
+use constant FLAVOURS => qw(linux_sysv linux_systemd solaris);
```
ned21: Since this is being exported can we switch to ReadOnly please? Otherwise changing this later means changing all the downstream dependencies.

jrha:
LGTM (as far as I can tell)

jrha:
@stdweird can you expand the PR description please?

jrha:
@ned21 happy?

ned21:
Will take a look over the weekend.

PR #119, opened 2015-11-24
Reduce verbosity of noaction mode

jrha:
Do not try to print the changes if the diff is empty. This makes the
output less noisy if a large number of files are opened, but only some
of them are changed.

Originally pushed to SF by @gombasg, but seems to have got lost on the way to GitHub.

### DIFF
--- a/src/main/perl/FileWriter.pm
+++ b/src/main/perl/FileWriter.pm
@@ -174,6 +174,8 @@ sub close
         if ($diff) {
             *$self->{LOG}->verbose ("Changes to ", *$self->{filename}, ":");
             *$self->{LOG}->report ($diff);
+        } else {
+            *$self->{LOG}->debug(1, "No changes to make to ", *$self->{filename});
         }
     }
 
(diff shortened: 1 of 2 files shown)


src/main/perl/FileWriter.pm
```diff
@@ -172,9 +172,10 @@ sub close
                                   stdin => "$self", stdout => \$diff,
                                   keeps_state => 1);
         $cmd->execute();
-        *$self->{LOG}->verbose ("Changes to ", *$self->{filename}, ":");
-        *$self->{LOG}->report ($diff);
-
+        if ($diff) {
+            *$self->{LOG}->verbose ("Changes to ", *$self->{filename}, ":");
+            *$self->{LOG}->report ($diff);
+        }
```
stdweird: should also do a log debug 1 tat nothing changed?

ned21:
LGTM. How do we see the unit test failure?

stdweird:
@jrha https://github.com/jrha/CAF/pull/1 (inlcudes resolved merge conflict with current master)
@ned21 the original unittests didn't do what they said they were doing

jrha:
@stdweird thanks, rebased for clarity.

PR #120, opened 2015-12-12
ObjectText: factored-out text related methods from TextRender 

stdweird:


src/main/perl/ObjectText.pm
```diff
@@ -0,0 +1,293 @@
...
+    }
+
+And use it via
+    my $sc = SubClass->new(log => $self);
+    print "$sc"; # stringification
+
+    $sc = SubClass->new(log => $self);
+    # return CAF::FileWriter instance (text already added)
+    my $fh = $sc->filewriter('/some/path');
+    if (!defined($fh)) {
```
stdweird: @ned21 what sort of error handling do you want to see here? simple failure or also catching `thorw_error`?

stdweird:
@ned21 ping?

src/main/perl/ObjectText.pm
ned21: I am not sure I understand the question sorry. What's the difference between the two approaches?

stdweird: creating/closing filewriter instance could throw an error, which is not catched in this example. but this code does not throw any errors itself, so as an example of how to use this module, this is sufficient.

jrha:
@ned21 any update?

jrha:
There is a whole chain of PRs based on this one, so I'm minded to merge it if there are no real concerns with it.

PR #121, opened 2015-12-23
Kerberos support via GSSAPI

stdweird:
Based on #120 

Important to note that kinit is not required at all.

### DIFF
--- a/src/main/perl/Kerberos.pm
+++ b/src/main/perl/Kerberos.pm
@@ -1,0 +1,763 @@
+# ${license-info}
+# ${developer-info}
+# ${author-info}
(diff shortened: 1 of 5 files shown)


stdweird:
GSSAPI is easy once you know how it works, but until then, you are stuck reading krb5-devel and the GSSAPI XS code, and lots of try and error :smile:

ned21:
Nothing obviously wrong jumps out at me but I'm not familiar at all with GSSAPI :(

jrha:
@stdweird rebase?

stdweird:
@jrha rebased

PR #128, opened 2016-01-05
Localize $@ when using eval

stdweird:
Fixes https://github.com/quattor/ncm-lib-blockdevices/issues/59


stdweird:
@jrha @ned21 scheduled for 16.2, but could be added to 15.12 (the issue it fixes is rare and is present in quattor since forever, so i see no real hurry)

ned21:
LGTM

PR #132, opened 2016-01-21
fix race condition in locking

ttyS4:
By introducing flock() a race condition can be avoided. This avoids if 2
ncm-ncd processes are started exactly the same time and both believe he
has the lock.

### DIFF
--- a/src/main/perl/Lock.pm
+++ b/src/main/perl/Lock.pm
@@ -9,10 +9,9 @@ use strict;
 use CAF::Object;
 use CAF::Reporter;
 
-use LC::Exception qw (SUCCESS throw_error);
+use LC::Exception qw(SUCCESS throw_error);
 use FileHandle;
 use Fcntl qw(:flock);
-#use Proc::ProcessTable;
 
 use Exporter;
 use vars qw(@ISA @EXPORT @EXPORT_OK);
@@ -22,9 +21,10 @@ use vars qw(@ISA @EXPORT @EXPORT_OK);
 @EXPORT_OK = qw(FORCE_ALWAYS FORCE_IF_STALE);
 
 
-use constant FORCE_NONE => 0;
-use constant FORCE_ALWAYS => 1;
-use constant FORCE_IF_STALE => 2;
+use constant FORCE_NONE     => 0;
+use constant FORCE_ALWAYS   => 1;
+use constant FORCE_IF_STALE => 2;  # for backwards compatibility only
+                                   # has no effect now
 
 
 =pod
@@ -35,19 +35,15 @@ CAF::Lock - Class for handling application instance locking
 
 =head1 SYNOPSIS
 
-
   use CAF::Lock;
 
-  $lock=CAF::Lock->new('/var/lock/quattor/spma');
-
-  if ($lock->is_locked()) {...} else {...};
+  $lock = CAF::Lock->new('/var/lock/quattor/spma');
 
   unless ($lock->set_lock()) {...}
-  unless ($lock->set_lock(10,2) {...}
-  unless ($lock->set_lock(3,3,FORCE_IF_STALE)) {...}
+  unless ($lock->set_lock(10, 2) {...}
+  unless ($lock->set_lock(3, 3, FORCE_ALWAYS)) {...}
 
   unless ($lock->unlock()) {....}
-
 
 =head1 INHERITANCE
 
@@ -56,7 +52,6 @@ CAF::Lock - Class for handling application instance locking
 =head1 DESCRIPTION
 
 The B<CAF::Lock> class provides methods for handling application locking.
-
 
 =over
 
@@ -70,7 +65,7 @@ The B<CAF::Lock> class provides methods for handling application locking.
 
 =back
 
-=head2 Public methods
+=head1 PUBLIC METHODS
 
 =over 4
 
@@ -93,28 +88,25 @@ sub is_locked {
 
 =item set_lock ($retries,$timeout,$force);
 
-Tries $retries times to set the lock. If $force is set to FORCE_NONE
+Tries I<retries> times to set the lock.  If I<force> is set to B<FORCE_NONE>
 or not defined and the lock is set, it sleeps for
-rand($timeout). Writes the current PID ($$) into the lock
-file. Returns SUCCESS or undef on failure.
+rand(I<timeout>).  Returns B<SUCCESS>, or B<undef> on failure.
 
-If $retries or $timeout are not defined or set to 0, only a single
+If I<retries> or I<timeout> are not defined or set to 0, only a single
 attempt is done to acquire the lock.
 
-If $force is set to FORCE_ALWAYS then the lockfile is just set
-again, independently if the lock is already set by another application
-instance.
-
-If $force is set to FORCE_ALWAYS then neither $timeout nor $retries are taken
+If I<force> is set to B<FORCE_ALWAYS> then the lock file is just set
+again, even if the lock is already set by another application
+instance, and neither I<timeout> nor I<retries> are taken
 into account.
 
 =cut
 
 sub set_lock {
-  my ($self,$retries,$timeout,$force)=@_;
+  my ($self, $retries, $timeout, $force) = @_;
 
-  $retries=0 unless (defined $retries);
-  $timeout=0 unless (defined $retries);
+  $retries = 0 unless (defined $retries);
+  $timeout = 0 unless (defined $retries);
 
   if ($self->{LOCK_SET}) {
     # oops.
@@ -122,7 +114,7 @@ sub set_lock {
     return undef;
   }
 
-  my $tries=0;
+  my $tries = 0;
   my $lock;
   while(1) {
     $tries++;
@@ -172,24 +164,24 @@ sub try_lock {
 
 =pod
 
-=item unlock
+=item unlock()
 
-Releases the lock and returns SUCCESS. Reports an error and returns
-undef if the lock file cannot be released. If the object (application
-instance) does not hold the lockfile, an error is reported and undef
+Releases the lock and returns B<SUCCESS>.  Reports an error and returns
+B<undef> if the lock cannot be released.  If the object (application
+instance) does not hold the lock, an error is reported and B<undef>
 is returned.
 
 =cut
 
 sub unlock {
-  my $self=shift;
+  my $self = shift;
   if ($self->{LOCK_SET}) {
     # if we forced the lock LOCK_FH can be undef
     return SUCCESS unless ($self->{LOCK_FH});
     unless ($self->{LOCK_FH}->close) {
       $self->error("cannot close lock file: ",$self->{'LOCK_FILE'});
     }
-    $self->{LOCK_SET}=undef;
+    $self->{LOCK_SET} = undef;
   } else {
     $self->error("lock not held by this application instance, not unlocking");
     return undef;
@@ -200,9 +192,9 @@ sub unlock {
 
 =pod
 
-=item is_set
+=item is_set()
 
-Returns SUCCESS if lock is set by application instance, undef otherwise
+Returns B<SUCCESS> if lock is set by application instance, B<undef> otherwise.
 
 =cut
 
@@ -218,11 +210,11 @@ sub is_set {
 
 =back
 
-=head2 Private methods
+=head1 PRIVATE METHODS
 
 =over 4
 
-=item _initialize($lockfilename)
+=item _initialize(I<lockfilename>)
 
 initialize the object. Called by new($lockfilename).
 
@@ -231,8 +223,8 @@ initialize the object. Called by new($lockfilename).
 sub _initialize {
   my ($self,$lockfilename) = @_;
 
-  $self->{'LOCK_SET'}=undef;
-  $self->{'LOCK_FILE'} = $lockfilename;
+  $self->{LOCK_SET} = undef;
+  $self->{LOCK_FILE} = $lockfilename;
   return SUCCESS;
 
 }
--- a/src/main/perl/Lock.pm
+++ b/src/main/perl/Lock.pm
@@ -72,8 +72,8 @@ The B<CAF::Lock> class provides methods for handling application locking.
 =item set_lock(I<retries>, I<timeout>, I<force>)
 
 Tries I<retries> times to set the lock.  If I<force> is set to B<FORCE_NONE>
-or not defined and the lock is set, it sleeps for
-rand(I<timeout>).  Returns B<SUCCESS>, or B<undef> on failure.
+or not defined and the lock is set, it sleeps for I<timeout>.  Returns
+B<SUCCESS>, or B<undef> on failure.
 
 If I<retries> or I<timeout> are not defined or set to 0, only a single
 attempt is done to acquire the lock.
@@ -101,7 +101,7 @@ sub set_lock {
   do {
     if ($tries > 0) {
       $self->verbose("lock file is already held, try $tries out of $retries");
-      sleep(rand($timeout));
+      sleep($timeout);
     }
     $tries++;
     return SUCCESS if $self->_try_lock($force);
--- a/src/main/perl/Lock.pm
+++ b/src/main/perl/Lock.pm
@@ -1,11 +1,5 @@
-# ${license-info}
-# ${developer-info}
-# ${author-info}
-# ${build-info}
+#${PMpre} CAF::Lock${PMpost}
 
-package CAF::Lock;
-
-use strict;
 use CAF::Object;
 use CAF::Reporter;
 
@@ -53,18 +47,6 @@ CAF::Lock - Class for handling application instance locking
 
 The B<CAF::Lock> class provides methods for handling application locking.
 
-=over
-
-=cut
-
-#------------------------------------------------------------
-#                      Public Methods/Functions
-#------------------------------------------------------------
-
-=pod
-
-=back
-
 =head1 PUBLIC METHODS
 
 =over 4
@@ -85,7 +67,8 @@ into account.
 
 =cut
 
-sub set_lock {
+sub set_lock
+{
     my ($self, $retries, $timeout, $force) = @_;
 
     $retries = 0 unless (defined $retries);
@@ -107,11 +90,9 @@ sub set_lock {
         return SUCCESS if $self->_try_lock($force);
     } while ($tries < $retries && $timeout);
 
-    $self->error("cannot acquire lock: $self->{LOCK_FILE}");
+    $self->error("cannot acquire lock after $tries tries: $self->{LOCK_FILE}");
     return;
 }
-
-=pod
 
 =item unlock()
 
@@ -122,17 +103,18 @@ is returned.
 
 =cut
 
-sub unlock {
+sub unlock
+{
     my $self = shift;
     if ($self->{LOCK_SET}) {
         # if we forced the lock LOCK_FH can be undef
         if ($self->{LOCK_FH}) {
             unless (flock($self->{LOCK_FH}, LOCK_UN)) {
-            $self->error("cannot release lock: $self->{LOCK_FILE}");
-            return;
-        }
-        $self->error("cannot close lock file: $self->{LOCK_FILE}")
-            unless $self->{LOCK_FH}->close();
+                $self->error("cannot release lock: $self->{LOCK_FILE}");
+                return;
+            }
+            $self->error("cannot close lock file: $self->{LOCK_FILE}")
+                unless $self->{LOCK_FH}->close();
         }
         $self->{LOCK_SET} = undef;
         $self->{LOCK_FH} = undef;
@@ -144,21 +126,17 @@ sub unlock {
 }
 
 
-=pod
-
 =item is_set()
 
 Returns B<SUCCESS> if lock is set by application instance, B<undef> otherwise.
 
 =cut
 
-sub is_set {
-    my $self=shift;
-    return SUCCESS if ($self->{LOCK_SET});
-    return;
+sub is_set
+{
+    my $self = shift;
+    return $self->{LOCK_SET} ? SUCCESS : ();
 }
-
-=pod
 
 =back
 
@@ -172,8 +150,9 @@ Initialize the object.  Called by new(I<lockfilename>).
 
 =cut
 
-sub _initialize {
-    my ($self,$lockfilename) = @_;
+sub _initialize
+{
+    my ($self, $lockfilename) = @_;
 
     $self->{LOCK_SET} = undef;
     $self->{LOCK_FILE} = $lockfilename;
@@ -181,7 +160,6 @@ sub _initialize {
     return SUCCESS;
 }
 
-=pod
 
 =item _try_lock(I<force>)
 
@@ -193,16 +171,18 @@ even if flock() was unsuccessful.
 
 =cut
 
-sub _try_lock {
+sub _try_lock
+{
     my ($self, $force) = @_;
-    my $lf = FileHandle->new("> " . $self->{LOCK_FILE});
+
+    my $lf = FileHandle->new("> $self->{LOCK_FILE}");
     unless ($lf) {
-        $self->error("cannot create lock file: " . $self->{LOCK_FILE});
+        $self->error("cannot create lock file: $self->{LOCK_FILE}");
         return;
     }
     unless (flock($lf, LOCK_EX|LOCK_NB)) {
         # Could not get the lock
-        return unless $force == FORCE_ALWAYS;
+        return unless (defined($force) && $force == FORCE_ALWAYS);
 
         # In force mode, continue but don't save the filehandle
         $lf->close();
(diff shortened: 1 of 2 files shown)


ttyS4:
Please test this.

stdweird:
test this please

stdweird:
test this please

msmark:
If the portability issues of `flock()` are not of concern, then I agree that using `flock()` would be a better approach.

src/main/perl/Lock.pm
```diff
@@ -85,7 +86,10 @@ If a lock is set for the lock file, returns SUCCESS, undef otherwise.
 
 sub is_locked() {
   my $self=shift;
-  return SUCCESS if (-e $self->{'LOCK_FILE'});
+  if (-e $self->{'LOCK_FILE'}) {
+    my $fh=FileHandle->new("> ". $self->{'LOCK_FILE'});
+    return SUCCESS if (not flock($fh, LOCK_EX|LOCK_NB));
```
msmark: You are locking the file momentarily in order to test that it is locked, then throwing the lock away again.  This is introducing a new race condition: a process could easily pass the first time `is_locked()` is called but fail the second time, causing the program to exit with an error.

Although you could use `fcntl` with `F_GETLK` to test for a lock without actually obtaining one, this is of dubious use as the lock could be gone before the `is_locked()` method returns.

It would be better to throw away the `is_locked()` method completely, and just have `set_lock()` attempt to run `flock()` multiple times with `LOCK_EX|LOCK_NB` until it wants to give up waiting any longer.

src/main/perl/Lock.pm
```diff
@@ -230,6 +236,8 @@ is returned.
 sub unlock {
   my $self=shift;
   if ($self->{LOCK_SET}) {
+    flock ($self->{LOCK_FH}, LOCK_UN|LOCK_NB) or $self->error("cannot release flock on lock file: ",$self->{'LOCK_FILE'});
+    $self->{LOCK_FH}=undef;
```
msmark: It would be preferable to `$self->{LOCK_FH}->close()` instead of `$self->{LOCK_FH}=undef`, as the intent is more explicit.  You should also test return status from the `close()` and report an error if you cannot close the lock file.

src/main/perl/Lock.pm
```diff
@@ -202,15 +206,17 @@ sub set_lock {
...
   my $lf=FileHandle->new("> " . $self->{'LOCK_FILE'});
+  $lf->autoflush;
   unless ($lf) {
     $self->error("cannot create lock file: ".$self->{'LOCK_FILE'});
     return undef;
   }
-  print $lf $$;
-  unless ($lf->close()) {
-    $self->error("cannot close lock file: ".$self->{'LOCK_FILE'});
+  unless (flock ($lf, LOCK_EX|LOCK_NB)) {
```
msmark: This is ignoring `FORCE_ALWAYS` and `FORCE_IF_STALE`, which means those options will be broken by this commit at present.

ttyS4:
True, this was a quick and minimal patch to tackle this issue:
. on flock portability - this should not be a concern based on http://perldoc.perl.org/perlport.html (and we are on a local filesystem)
. reporting error on fileclose - yes, i can do that however from the locking perspective it does not really matter (also see below on this)

I agree; with more restructuring this could play by the rules trying to re-acquire the lock if it is taken and a stale lock could be detected if we find a file without flock (at this point the content of the file the pid number is not needed anymore).

Mark, do you agree?

msmark:
The `flock()` portability question is one for other Quattor maintainers.  For our use case it works on all the OS's we care about, but I was flagging it so that others can comment.

Reporting error on file close does matter, it helps to troubleshoot problems if there is some sort of I/O error.

No you don't want to try to re-acquire the lock if it is taken.  It can't be stolen away from you.  I see no need to detect a file that isn't locked either.  Keep it simple - the purpose is to ensure that only one `ncm-ncd` process is running at any given time.

jrha:
AFAIK `flock` is implemented on all Perl ports other than VMS, RISC OS and VOS, so I think we're probably fine.

ttyS4:
I fixed most issues reported by @msmark.

I even dropped get_lock_pid() after confirming with him.
Some remained:
is_locked() - shows the current state which could have some value. If the lock is taken the real lock attempt could fail later. Still it is not used inside Lock.pm nor ncm-ncd so may it could be dropped.
try_lock() - it still reports errors multiple times (they may not be all the same) and I do not see it as a big deal.

msmark:
Try this: https://github.com/ttyS4/CAF/pull/1

stdweird:
test this please

gombasg:
Perl's flock() is suboptimal, because its implementation depends on how Perl was compiled. If there are multiple Perl interpreters available, and they were built with different options, then it's still possible for multiple processes to own the same lock simultaneously.

There are some other possibilities:

- Use fcntl() instead of flock(). fcntl() has the most chance of working, however it requires dealing with architecture-specific packed structures, which is hard to do properly from pure Perl.

- Use O_EXCL. O_EXCL should work for most POSIX-like file systems; it can break e.g. if you export a filesystem having CIFS-like permissions over NFS, but I think that would be acceptable.

In either case, it would be good if the location of the lock file could be set independent from the location of the file to be locked (that may already be the case, I did not check), so if the file to be protected happens to live on some really strange filesystem, the lockfile can still be put at a saner location.

ttyS4:
@gombasg Yes, what you are saying is valid, but not really a new thing.

This implementation requires that the perl interpreter was built with the same locking options (this seems reasonable to me).
I agree that fcntl() is not feasible.

I did not get your other suggestion, please elaborate.
The lockfile location is not hard-coded.

ttyS4:
Today I included the patch by @msmark. Thanks for that Mark and sorry because I was long overdue with this one.
I did some tests and all seems fine to me.

stdweird:
@ttyS4 can you make a request to join the quattor developers group https://github.com/orgs/quattor/teams/developers
that should enable the jenkins testing for your PRs

ttyS4:
@stdweird I tried and that link does not work for me (404).

msmark:
That's funny, the link works for me.

stdweird:
@msmark you are already in the group, that's maybe the reason.
@ttyS4 we will try to add you

ned21:
I've invited @ttyS4 to the organisation and the team. I think the problem was he was not in the Quattor org, so couldn't join the developers team.

ttyS4:
Works now, thx @ned21

ttyS4:
test this please

src/main/perl/Lock.pm
```diff
@@ -246,63 +146,90 @@ sub unlock {
...
 sub is_set {
   my $self=shift;
   return SUCCESS if ($self->{LOCK_SET});
-  return undef;
+  return;
 }
 
+##############################################################################
+#
+# The following methods are provided for backwards compatibility only
```
stdweird: this is not what i consider `backwards compatibility`. these are noops that do something very very different from the original code.
i'd rather have you remove these then to keep them around implemented like this. if you want to keep them around, at least have them `->error('Danger danger: contact developers')` or something very obvious.

i grepped the quattor code, these 3 methods are only used in `CAF::Lock`, i vote to remove them.

src/main/perl/Lock.pm
```diff
@@ -246,63 +146,90 @@ sub unlock {
...
-=item is_set
+=item is_set()
 
-Returns SUCCESS if lock is set by application instance, undef otherwise
+Returns B<SUCCESS> if lock is set by application instance, B<undef> otherwise.
 
 =cut
 
 sub is_set {
   my $self=shift;
```
stdweird:
```perl
my $self = shift;
return $self->{LOCK_SET} ? SUCCESS : undef;
```

src/main/perl/Lock.pm
```diff
@@ -73,172 +65,80 @@ The B<CAF::Lock> class provides methods for handling application locking.
...
-	   ($self->is_locked() && !($force == FORCE_IF_STALE && $self->is_stale()))) {
-      my $sleep=rand($timeout);
-      $self->verbose("lockfile is already held, try $tries out of $retries");
-      $tries++;
-      sleep($sleep);
+  my $tries = 0;
+  do {
+    if ($tries > 0) {
+      $self->verbose("lock file is already held, try $tries out of $retries");
+      sleep(rand($timeout));
```
stdweird: why the `rand`?

src/main/perl/Lock.pm
```diff
@@ -9,21 +9,22 @@ use strict;
...
-@EXPORT_OK = qw(FORCE_ALWAYS FORCE_IF_STALE);
+@EXPORT_OK = qw(FORCE_NONE FORCE_ALWAYS FORCE_IF_STALE);
 
 
-use constant FORCE_NONE => 0;
-use constant FORCE_ALWAYS => 1;
-use constant FORCE_IF_STALE => 2;
+use constant FORCE_NONE     => 0;
+use constant FORCE_ALWAYS   => 1;
+use constant FORCE_IF_STALE => 2;  # for backwards compatibility only
```
stdweird: if this value is used somewhere, at least report it with `verbose` or `info`, and mention that it has no effect and that developers should be contacted.

src/main/perl/Lock.pm
```diff
@@ -73,172 +65,80 @@ The B<CAF::Lock> class provides methods for handling application locking.
...
+  my ($self, $retries, $timeout, $force) = @_;
 
-  $retries=0 unless (defined $retries);
-  $timeout=0 unless (defined $retries);
+  $retries = 0 unless (defined $retries);
+  $timeout = 0 unless (defined $retries);
 
   if ($self->{LOCK_SET}) {
     # oops.
     $self->error("lock already set by this application instance");
```
stdweird: all messages, esp the error ones, are not that useful without the lockfile. just add a `$self->{LOCK_FILE}` in the double-quoted message (no need for `".$self->`). e.g. this message could become
```perl
$self->error("lock already set by this application instance: $self->{LOCK_FILE}");
```

src/main/perl/Lock.pm
```diff
@@ -73,172 +65,80 @@ The B<CAF::Lock> class provides methods for handling application locking.
...
-    while ($tries <= $retries &&
-	   ($self->is_locked() && !($force == FORCE_IF_STALE && $self->is_stale()))) {
-      my $sleep=rand($timeout);
-      $self->verbose("lockfile is already held, try $tries out of $retries");
-      $tries++;
-      sleep($sleep);
+  my $tries = 0;
+  do {
+    if ($tries > 0) {
+      $self->verbose("lock file is already held, try $tries out of $retries");
```
stdweird: add LOCK_FILE to message

src/main/perl/Lock.pm
```diff
@@ -73,172 +65,80 @@ The B<CAF::Lock> class provides methods for handling application locking.
...
-    return undef;
-  }
-  $self->{LOCK_SET}=1;
-  return SUCCESS;
-}
+    $tries++;
+    return SUCCESS if $self->_try_lock($force);
+  } while ($tries < $retries && $timeout);
 
+  $self->error("cannot acquire lock: " . $self->{LOCK_FILE});
```
stdweird: no need for the `" . $self->`, add it in the double quoted text

src/main/perl/Lock.pm
```diff
@@ -73,172 +65,80 @@ The B<CAF::Lock> class provides methods for handling application locking.
...
+        $self->error("cannot release lock: ",$self->{LOCK_FILE});
+        return;
+      }
+      $self->error("cannot close lock file: ",$self->{LOCK_FILE})
+          unless $self->{LOCK_FH}->close();
     }
+    $self->{LOCK_SET} = undef;
+    $self->{LOCK_FH} = undef;
   } else {
     $self->error("lock not held by this application instance, not unlocking");
```
stdweird: add LOCK_FILE to message

src/main/perl/Lock.pm
```diff
@@ -73,172 +65,80 @@ The B<CAF::Lock> class provides methods for handling application locking.
...
   if ($self->{LOCK_SET}) {
-    unless (unlink($self->{'LOCK_FILE'})) {
-      $self->error("cannot release lock file: ",$self->{'LOCK_FILE'});
-      return undef;
-    } else {
-      $self->{LOCK_SET}=undef;
+    # if we forced the lock LOCK_FH can be undef
+    if ($self->{LOCK_FH}) {
+      unless (flock($self->{LOCK_FH}, LOCK_UN)) {
+        $self->error("cannot release lock: ",$self->{LOCK_FILE});
```
stdweird: add the `$self->{LOCK_FILE}` to the string

src/test/perl/lock.t
```diff
@@ -13,38 +13,10 @@ unlink(LOCK_TEST);
 my $lock=CAF::Lock->new(LOCK_TEST);
 
 ok(!$lock->is_locked(), "Unlocked at start");
-my $lockpid=$lock->get_lock_pid();
-
-is($lockpid, undef, "Lock PID undefined on unaquired lock");
 
 ok($lock->set_lock(), "Lock set");
 ok($lock->is_locked(), "Locked on request");
```
stdweird: add a new `CAF::Lock` to try to take the lock on the same file. 
should fail after this (and is an actual test to see if the locking works).

src/test/perl/lock.t
```diff
@@ -13,38 +13,10 @@ unlink(LOCK_TEST);
...
 
-is($lock->get_lock_pid(), $$, "Lock PID correctly set on locked object");
-
-ok(!$lock->is_stale(), "Lock is NOT stale");
-
 ok($lock->unlock(), "Lock released");
```
stdweird: take the lock with the seconds lock instances, it should work now (and that is the actual test that the lock was released)

stdweird:
@ttyS4 thanks for cleaning this up. i would really like you to improve the unittests by adding a second `CAF::Lock` instance and try to take the lock.
also, our coding convention is 4 space indentation, and as you can see, this code have seen much caretaking in a long while :cry: )
once the remarks are addressed, it would be really nice if you made a final seprate commit to fix all whitespace indentation.

src/main/perl/Lock.pm
```diff
@@ -9,21 +9,22 @@ use strict;
...
-use LC::Exception qw (SUCCESS throw_error);
+use LC::Exception qw(SUCCESS throw_error);
 use FileHandle;
-#use Proc::ProcessTable;
+use Fcntl qw(:flock);
 
 use Exporter;
 use vars qw(@ISA @EXPORT @EXPORT_OK);
 
 @ISA = qw(CAF::Reporter CAF::Object Exporter);
```
stdweird: There's something seriously wrong with this. `CAF::Lock` is it's own reporter? i opened #146 to follow up as soon as this is merged.

src/main/perl/Lock.pm
ttyS4: for one: it was always like that.
two: it could make sense to do this in a distributed environment.

Also if there are multiple processes waiting for some resource that is protected by the lock randomization makes sense (to me).

ttyS4:
@stdweird Do you have some perltidy policy documented somewhere?
I would prefer to run it thru perltidy instead of doing manually.

src/main/perl/Lock.pm
stdweird: i'm not sure it make any difference, except that it lowers the odds that you actually get the lock before you run out of retries (in praticular, the average time to wait before giving up is `timeout * tries /2`).
if you are trying to deal with 2 threads started at exactly the same time, i think it make more sense to do
```perl
sleep($timeout -1 + rand(2))
```
(sleep works with integer seconds, this will give some fluctuation; if that would matter at all)
but i wasn't aware sleeping was a race; only one lock can be held at any given time anyway.

stdweird:
@ttyS4 no real policy, i used to use the following for while
```
perltidy --indent-columns=4 --maximum-line-length=100 --continuation-indentation=4 --check-syntax --converge --nooutdent-long-lines --paren-tightness=2 --square-bracket-tightness=2 --brace-tightness=2 --block-brace-tightness=2 --nospace-for-semicolon --nospace-function-paren --cuddled-else --noopening-brace-on-new-line --opening-sub-brace-on-new-line
```

but now i clean up by hand while reviewing all other outdated code :smile: 
the important part if to make a single commit with all whitespace changes

ned21:
Is this OK to be merged for 16.4? Does @ttyS4 need to rebase first?

ttyS4:
I will take a look late today.

On May 5, 2016 4:31:26 PM GMT+02:00, Nathan Dimmock <notifications@github.com> wrote:
>Is this OK to be merged for 16.4? Does @ttyS4 need to rebase first?
>
>---
>You are receiving this because you were mentioned.
>Reply to this email directly or view it on GitHub:
>https://github.com/quattor/CAF/pull/132#issuecomment-217169763

-- 
Csillag Tamás 
mobilról

ttyS4:
@stdweird FORCE_IF_STALE is still used here: https://github.com/quattor/CCM/blob/master/src/main/perl/Fetch/ProfileCache.pm#L50

ttyS4:
with FORCE_IF_STALE added back (as a noop) the tests succeeds
The log messages from the failed build can be used to find some users of this option.

I tried to address the previous questions. I will do the reformatting next.

msmark:
We should fix ProfileCache to remove the 3rd argument sent to `set_lock()` so that it no longer needs to reference `FORCE_IF_STALE`, which is now redundant.

ttyS4:
@msmark on one hand I agree with you on the other hand I consider this mergeable.
These issues are hygiene and can be dealt in the background. Once they have been fixed the compat code can be removed.

If we merge this I can open a follow-up pull request to track this if you want.
(Oh, and I am doing the non-code re-indent now.)

src/main/perl/Lock.pm
```diff
@@ -73,237 +65,155 @@ The B<CAF::Lock> class provides methods for handling application locking.
...
-      $tries++;
-      sleep($sleep);
+    my ($self, $retries, $timeout, $force) = @_;
+
+    $retries = 0 unless (defined $retries);
+    $timeout = 0 unless (defined $retries);
+
+    if ($self->{LOCK_SET}) {
+        # oops.
+        $self->error("lock already set by this application instance: $self->{LOCK_FILE}");
```
stdweird: why is this an error? a coding mistake? 
afai can see it, you try to set the lock, and the end result is that you have the lock (because you already had it).

src/main/perl/Lock.pm
```diff
@@ -73,237 +65,155 @@ The B<CAF::Lock> class provides methods for handling application locking.
...
+    do {
+        if ($tries > 0) {
+            $self->verbose("lock file is already held, try $tries out of $retries");
+            sleep($timeout);
+        }
+        $tries++;
+        return SUCCESS if $self->_try_lock($force);
+    } while ($tries < $retries && $timeout);
+
+    $self->error("cannot acquire lock: $self->{LOCK_FILE}");
```
stdweird: add number of attempts to the message

src/main/perl/Lock.pm
```diff
@@ -73,237 +65,155 @@ The B<CAF::Lock> class provides methods for handling application locking.
...
 sub unlock {
-  my $self=shift;
-  if ($self->{LOCK_SET}) {
-    unless (unlink($self->{'LOCK_FILE'})) {
-      $self->error("cannot release lock file: ",$self->{'LOCK_FILE'});
-      return undef;
+    my $self = shift;
+    if ($self->{LOCK_SET}) {
+        # if we forced the lock LOCK_FH can be undef
+        if ($self->{LOCK_FH}) {
```
stdweird: if `$self->{LOCK_FH}` can be 0 or `''`, this is not going to work. better use `if (defined($self->{LOCK_FH}))`

src/main/perl/Lock.pm
```diff
@@ -73,237 +65,155 @@ The B<CAF::Lock> class provides methods for handling application locking.
...
+            unless (flock($self->{LOCK_FH}, LOCK_UN)) {
```
stdweird: the indentation here and next 5 lines is wrong

src/main/perl/Lock.pm
```diff
@@ -73,237 +65,155 @@ The B<CAF::Lock> class provides methods for handling application locking.
...
+            $self->error("cannot release lock: $self->{LOCK_FILE}");
+            return;
+        }
+        $self->error("cannot close lock file: $self->{LOCK_FILE}")
```
stdweird: no `return` here?

src/main/perl/Lock.pm
```diff
@@ -73,237 +65,155 @@ The B<CAF::Lock> class provides methods for handling application locking.
...
-Returns SUCCESS if lock is set by application instance, undef otherwise
+Returns B<SUCCESS> if lock is set by application instance, B<undef> otherwise.
 
 =cut
 
 sub is_set {
-  my $self=shift;
-  return SUCCESS if ($self->{LOCK_SET});
-  return undef;
+    my $self=shift;
```
stdweird: spaces around `=`

src/main/perl/Lock.pm
```diff
@@ -73,237 +65,155 @@ The B<CAF::Lock> class provides methods for handling application locking.
...
+    return SUCCESS if ($self->{LOCK_SET});
```
stdweird: use
```perl
return $self->{LOCK_SET} ? SUCCESS : undef;
```

src/main/perl/Lock.pm
stdweird: no need, this appears to a class instance.

src/test/perl/lock.t
```diff
@@ -2,49 +2,26 @@
...
 unlink(LOCK_TEST);
 
-my $lock=CAF::Lock->new(LOCK_TEST);
+my $lock1 = CAF::Lock->new(LOCK_TEST);
+my $lock2 = CAF::Lock->new(LOCK_TEST);
 
-ok(!$lock->is_locked(), "Unlocked at start");
-my $lockpid=$lock->get_lock_pid();
+ok(!$lock1->is_set(), "Unlocked at start");
+ok($lock1->set_lock(), "Lock set");
```
stdweird: add test for a 2nd `$lock1->set_lock()`

src/test/perl/lock.t
```diff
@@ -2,49 +2,26 @@
...
 mkdir(LOCK_TEST_DIR);
 unlink(LOCK_TEST);
 
-my $lock=CAF::Lock->new(LOCK_TEST);
+my $lock1 = CAF::Lock->new(LOCK_TEST);
+my $lock2 = CAF::Lock->new(LOCK_TEST);
 
-ok(!$lock->is_locked(), "Unlocked at start");
-my $lockpid=$lock->get_lock_pid();
+ok(!$lock1->is_set(), "Unlocked at start");
```
stdweird: add same test here for `lock2`.

src/test/perl/lock.t
```diff
@@ -2,49 +2,26 @@
...
-$lock->unlock();
-
-open($fh, ">", LOCK_TEST);
-close($fh);
-
-ok($lock->is_stale(), "Lock on empty lock file is stale");
-ok($lock->set_lock(0, 0, FORCE_IF_STALE), "Forced lock on empty lock file aquired");
-
-# How does it handle when an empty lock file is there?
+ok(!$lock2->set_lock(), "Second lock when file locked");
```
stdweird: add a `is_set` test here

src/test/perl/lock.t
```diff
@@ -2,49 +2,26 @@
...
 
+ok($lock1->unlock(), "Lock released");
```
stdweird: add a `lock1->is_set` test here

src/test/perl/lock.t
```diff
@@ -2,49 +2,26 @@
...
 
+ok($lock2->set_lock(), "Second lock when file unlocked");
+ok($lock2->is_set(), "Second lock locked on request");
```
stdweird: add a `lock1->is_set` test here

src/test/perl/lock.t
stdweird: the main reason for these extra tests is to make sure that instances do not share any states. if lock1 takes the lock, lock2 should be unmodified

stdweird:
@ttyS4 https://github.com/ttyS4/CAF/pull/2 fixes my remarks, except 2 things:
`set_lock` when lock is already taken should not give an error imho. same for `unkock` when no lock is set.
(the unittests in my PR comply with current code, so those will have to be changed as well)
@ned21 toughts on the behaviour of set_lock when the instacne already has the lock?

ned21:
@stdweird Taking a lock should be idempotent so downgrade the message to a warning() and return SUCCESS.

stdweird:
@ned21 fix in last commit of https://github.com/ttyS4/CAF/pull/2
once that PR is merged, this is ready to be merge in master

ttyS4:
Sorry, i was busy recently. Back again with adopting the changes suggested.
Hopefully we are all set :-)

stdweird:
@ttyS4 thanks for attacking this
@jrha @jouvin LGTM, but i'm not going to merge my own commits :smile:

PR #140, opened 2016-02-16
FileReader: fix example

### LINKED ISSUE #139: CAF::FileReader example use does not work
CAF::FileReader includes the following sample code but [a search ](https://github.com/quattor/configuration-modules-core/search?utf8=%E2%9C%93&q=FileReader&type=Code) of the code base shows no actual examples of its use.

```
    use CAF::FileReader;
    my $fh = CAF::FileReader->open ("my/path");
    while (my $line = <$fh>) {
       # Do something
    }
```

In testing, this doesn't work. ```print "$fh"``` does work because of stringify().  This works:

```
my $fh = CAF::FileReader->open("/path/to/file");
foreach my $line (split(/\n/, "$fh")) {
    print "$line\n";
}
```

Unclear if this is a documentation bug or if filehandle syntax should work?

stdweird:
Fixes #139

### DIFF
--- a/src/main/perl/FileReader.pm
+++ b/src/main/perl/FileReader.pm
@@ -18,7 +18,7 @@ CAF::FileReader - Class for only reading files in CAF applications.
 Normal use:
 
     use CAF::FileReader;
-    my $fh = CAF::FileReader->open ("my/path");
+    my $fh = CAF::FileReader->new ("my/path");
     while (my $line = <$fh>) {
        # Do something
     }


stdweird:
@ned21 open also does not cancel, giving you an editor rather than a reader

ned21:
Can you explain why open != new given:

```
# Alias open to new.
no warnings 'redefine';
*__PACKAGE__::open = \&new;
use warnings;
```

stdweird:
you are right, open is an alias for new in Writer and Editor too. i'll have a look

PR #141, opened 2016-02-16
Correct aliasing of new/open for CAF::FileWriter/Editor/Reader

### LINKED ISSUE #139: CAF::FileReader example use does not work

stdweird:
Fixes #139

### DIFF
--- a/src/main/perl/FileEditor.pm
+++ b/src/main/perl/FileEditor.pm
@@ -79,10 +79,11 @@ Synonym for C<new()>
 
 =cut
 
-sub open
-{
-    return new(@_);
-}
+# Alias open to new.
+no warnings 'redefine';
+*open = \&new;
+use warnings;
+
 
 =pod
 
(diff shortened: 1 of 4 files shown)


stdweird:
The unittest failed for original Writer and Reader, so expect some strange stuff.
Also 'requires perl guru' to know why the old code was wrong.
@jrha i'll leave the milestone up to you

ned21:
Assuming we want to continue to have open as a synonym for new this looks fine. I'd suggest 16.4 in case this introduces some subtle bugs that delay the release.

jrha:
@stdweird Github is telling me this needs to be rebased

stdweird:
and now it has

jrha:
Excellent

PR #142, opened 2016-03-06
Path: common file and directory operations similar to LC::Check

### LINKED ISSUE #69: CAF::FileStatus
Several components (see https://github.com/quattor/configuration-modules-grid/issues/58 for an example) rely on `LC::Check` features that are not implemented in `CAF::FileXxx`. This is mainly:
- `LC::Check::directory`: creation of a directory if it doesn't exist
- `LC::Check::status` for directories: allow to define owner/group/perm... for directories
- `LC::Check::absence`: remove a file/directory if it exists
- `LC::Check::symlink`: symlink management

Also `LC::Check::status` for file can be executed through `CAF::FileEditor` but this is a bit of an overkill to use this interface if you just want to check/update file metadata.

It may be useful to have a `CAF::FileStatus` module that will act as a wrapper over `LC::Check` or other standard Perl module/functions (e.g. `File::Path`) to implement high-level features that can be tested (mocked) as other `CAF` modules.

Also for directory, it may be interesting to offer a `directory` function that combines `LC::Check::directory` and `LC::Check::status` as this is not very efficient to have to call 2 functions... This would be more consistent with `CAF::FileXxx` modules.

stdweird:
Initial framework towards resolving #69
Changes include (but not limited to):
* catches all exceptions (no execption are raised, existing ones are ignored and forgottten)

### DIFF
--- a/src/main/perl/Check.pm
+++ b/src/main/perl/Check.pm
@@ -19,6 +19,7 @@ Readonly::Hash my %CLEANUP_DISPATCH => {
 # Use dispatch table instead of non-strict function by variable call
 Readonly::Hash my %LC_CHECK_DISPATCH => {
     directory => \&LC::Check::directory,
+    status => \&LC::Check::status,
 };
 
 our $EC = LC::Exception::Context->new->will_store_all;
@@ -317,6 +318,21 @@ sub directory
 {
     my ($self, $directory, %opts) = @_;
     return $self->LC_Check("directory", [$directory], \%opts);
+}
+
+=item status
+
+Set the path stat options: C<owner>, C<group>, C<mode> and/or C<mtime>.
+
+This is a wrapper around C<LC::Check::status>
+and executed with C<LC_Check>.
+
+=cut
+
+sub status
+{
+    my ($self, $path, %opts) = @_;
+    return $self->LC_Check("status", [$path], \%opts);
 }
 
 =pod
--- a/src/main/perl/Check.pm
+++ b/src/main/perl/Check.pm
@@ -1,14 +1,17 @@
 ${PMpre} CAF::Check${PMpost}
 
-use CAF::Object;
+use CAF::Object qw(SUCCESS CHANGED);
 use LC::Check;
-use LC::Exception qw (SUCCESS throw_error);
+use LC::Exception qw (throw_error);
+
 use Readonly;
 
 use File::Path qw(rmtree);
 use File::Copy qw(move);
 use File::Temp qw(tempdir);
 use File::Basename qw(dirname);
+
+use Scalar::Util qw(dualvar);
 
 Readonly my $KEEPS_STATE => 'keeps_state';
 
@@ -22,7 +25,6 @@ Readonly::Hash my %CLEANUP_DISPATCH => {
 Readonly::Hash my %LC_CHECK_DISPATCH => {
     directory => \&LC::Check::directory,
     status => \&LC::Check::status,
-    reset_exception => sub {return 1;}, # do nothing
 };
 
 our $EC = LC::Exception::Context->new->will_store_all;
@@ -59,6 +61,18 @@ The class is based on L<LC::Check> with following major difference
 undef on failure and store the error message in the C<fail> attribute.
 
 =item available as class-methods
+
+=item return values
+
+=over
+
+=item undef: failure occured
+
+=item SUCCESS: nothing changed (boolean true)
+
+=item CHANGED: something changed (boolean true).
+
+=back
 
 =back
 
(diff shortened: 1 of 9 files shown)


stdweird:
To discuss at workshop
* the name :smile: 
* exportable function (instead or together with) methods

stdweird:
I'll be adding more unittests, and also a PR to maven-tools to allow easy unittesting from Test::Quattor

src/main/perl/Check.pm
```diff
@@ -0,0 +1,321 @@
...
+
+=cut
+
+sub _function_catch
+{
+    my ($self, $funcref, $args, $opts) = @_;
+
+    # Reset the fail attribute
+    if ($self->{fail}) {
+        $self->verbose("Ignoring/resetting previous existing fail: ",
```
ned21: If this is something the user should care about then perhaps it should be warn? If it's for developers then debug()

src/main/perl/Check.pm
```diff
@@ -0,0 +1,321 @@
...
+sub _get_noaction
+{
+    my ($self, $keeps_state, $msg) = @_;
+
+    $msg = '' if (! defined($msg));
+
+    my $noaction;
+
+    if ($keeps_state) {
+        $self->verbose($msg, "keeps_state set, noaction is false");
```
ned21: verbose seems too high here for something that's not informative to a user. The user needs to see the normal log message modified to confirm that no action was actually taken, this sort of trace level message is only useful to developers and should go into debug()

ned21:
Thanks for putting this together, Stijn. I guess a meta-question is whether we want to wrap LC::Check in this way, given it is no longer supported in its current form, or whether we should just start over? Or is the intention to insert this shim, move all components to use it and then replace these functions with in place implementations over time. The latter option sounds pretty attractive given that the unit tests should make that refactoring fairly low risk.

stdweird:
@ned21 yes, my main motiviations are to create a path to move away from LC::Check and to be able to make unittesting easier. the `cleanup` method is a reimplementation because there was no code that handled backing up directories.

stdweird:
@ned21 this is now complete wrt tests.

next is handling `LC::Check::_message` / silent option, which will lead to similar discussion as in https://github.com/quattor/CAF/pull/98 (and the proposal to use `CAF::History`).
i'll make separate PR for that.

adding more LC::Check function is left as an exercise for those who need it and/or want to complete #69 (i'll be lazy here, i still need to add this framework in `Test::Quattor` :smile: )

stdweird:
See #143 for handling `_message`

jrha:
Did we discuss this at the workshop? I can't recall.

stdweird:
yes we did, it's briefly mentioned in the workshop notes

ned21:
What name did the workshop decide upon? CAF::Check is not ideal since it's not meaningful to anyone who doesn't know the history of LC::Check

stdweird:
@ned21 no alternative names were proposed. i had the impression that noone really objected, but than also everyone knew `LC::Check`. and i'm just bad at proposing names. if you have a better proposal, now's the time :smile:

ned21:
hmm... I'm not great at names but I'll give it a go. CAF::PathUtils ?

stdweird:
fine by me, i wouldn't mind naming it `CAF::Path` (`CAF` already implies `bunch of utils` for me, and it is a not a set of functions, but rather a set of additional methods).

stdweird:
@ned21 i partially rewrote it to enforce consistent return values. you might want to rereview the code.

src/main/perl/Check.pm
```diff
@@ -338,62 +424,95 @@ will be made at the end of the program).
 
 =cut
 
+# Differences with LC::Check::directory
+#    only accepts one directory
+#    returns directory name or undef (LC::Check::directory returns number of created directories)
```
ned21: returns a dualvar (containing the directory name) or undef

ned21:
Code LGTM. Just need to decide on a  name. CAF::Path is fine with me.

src/main/perl/Check.pm
```diff
@@ -0,0 +1,524 @@
...
+attempt (incl. any files and/or subdirectries)
+will be made at the end of the program).
+
+=back
+
+=cut
+
+# Differences with LC::Check::directory
+#    only accepts one directory
+#    returns directory name or undef (LC::Check::directory returns number of created directories)
```
ned21: Re-commenting here as GH seems to have lost it. 

returns a dualvar (containing the directory name) or undef

stdweird:
@jrha @jouvin any preferences wrt `CAF::Check`, `CAF::PathUtils` or `CAF::Path`. i'm only renaming once :smile:

stdweird:
renamed the whole thing to `CAF::Path`

jrha:
LGTM

PR #145, opened 2016-03-25
Kerberos: make it work

stdweird:
The previous CAF::Kerberos had the correct code path, but was not working because interacting with XS requires careful passing around of the variables.


stdweird:
This is now actually tested and works with e.g. kerberos authenticated ccm-fetch

jrha:
LGTM

stdweird:
@jrha @ned21 rebased, tests ok

PR #147, opened 2016-03-31
Reporter: add methods to query state of reporter

stdweird:
Based on #117


ned21:
Do this need to be rebased?

stdweird:
@ned21 i'm working on a component that uses external modules that supports regular and very verbose logging, and i want to turn on the verbose logging of that module when the component runs with `--debug 3` or higher. previously there was no clean way to obtain the current log level.

PR #154, opened 2016-04-24
FileWriter: add conditional logger interfaces

stdweird:
Requires new build-scripts with https://github.com/quattor/maven-tools/pull/80

### DIFF
--- a/src/main/perl/FileEditor.pm
+++ b/src/main/perl/FileEditor.pm
@@ -46,16 +46,16 @@ the class constructor.
 =item new
 
 Returns a new object it accepts the same arguments as the constructor
-for C<CAF::FileWriter> with one additional option: 
+for C<CAF::FileWriter> with one additional option:
 
 =over
 
-=item source 
+=item source
 
 This option, when present, must be a file name whose contents will be used
 as the initial contents for the edited file if the source modification time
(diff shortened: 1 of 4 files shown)


stdweird:
based on #155 which has the buildtools versionbump
tests pass again

jrha:
test this please

stdweird:
@jouvin incl the RuleBasedEditor

jrha:
LGTM (as far as I can tell)

ned21:
This just does ```*$self->{LOG}->debug(...)``` -> `$self-debug(...)```. That's nicer to look at but a longer explanation of why and how would be good!

jouvin:
As said (suggested!) in the initial description,  https://github.com/quattor/CAF/pull/154#issuecomment-218285819 defined the usual `error()`, `debug()` method in the test context where they were undefined for things other than a component (no reporter). This allows both a more readable code and also a simpler one as `$self->{LOG}` can be undefined and thus every `$self->{LOG}->xxx(...)` had to be followed by `if $self->{LOG}`.

stdweird:
@ned21 the conditional reporter interface is already part of `CAF::Object`, and as @jouvin pointed out, the main benefit is that you can get rid of the `if *$self->{LOG}` in the code.

PR #156, opened 2016-04-25
FileEditor: add a 'source' option in constructor

jouvin:
This PR adds to the FileEditor the capability to open a reference version of the file instead of the file itself, if the reference file happens to be newer than the file. The typical use case is a file provided by a RPM that you edit with Quattor. You may want that if a new version of the RPM installs a new file as `.rpm-new`, your next run of Quattor applies modifications to this new version (this is particularly useful with the RuleBasedEditor). 

The idea is that you do it by adding a 'source' option whose value is a file name. If the source doesn't exist, it is just ignored. In this case the file will be used if it exists (current behaviour).

### DIFF
--- a/src/main/perl/FileEditor.pm
+++ b/src/main/perl/FileEditor.pm
@@ -58,13 +58,38 @@ sub _is_valid_source
     return -f $fn;
 }
 
+sub _is_reference_newer
+{
+    my ($self) = @_;
+    my $is_newer = 0;   # Assume false
+    if ( $self->_is_valid_source(*$self->{options}->{source}) ) {
+        if ( !$self->_is_valid_source(*$self->{filename}) || 
+             ((stat(*$self->{options}->{source}))[9] > (stat(*$self->{filename}))[9]) ) {
+          $is_newer = 1
+        }
+    }
+    if ( $is_newer ) {
+        *$self->{LOG}->debug(2, "Source file (", *$self->{options}->{source}, " is newer than ", *$self->{filename}, ": use it");
+    } else {
+        *$self->{LOG}->debug(2, "Source file (", *$self->{options}->{source}, " older than ", *$self->{filename}, ": ignore it");
+    };
+    return $is_newer;
+}
 
 sub new
 {
     my $class = shift;
     my $self = $class->SUPER::new (@_);
-    if ($self->_is_valid_source(*$self->{filename})) {
-        my $txt = LC::File::file_contents (*$self->{filename});
+    my $txt;
+    my ($path, %opts) = @_;
+
+    *$self->{options}->{source} = $opts{source} if exists ($opts{source});
+    if ( exists(*$self->{options}->{source}) && $self->_is_reference_newer() ) {
+        $txt = LC::File::file_contents (*$self->{options}->{source});
+    } elsif ($self->_is_valid_source(*$self->{filename})) {
+        $txt = LC::File::file_contents (*$self->{filename});
+    }
+    if ( $txt ) {
         $self->IO::String::open ($txt);
         $self->seek(IO_SEEK_END);
     }
--- a/src/main/perl/FileEditor.pm
+++ b/src/main/perl/FileEditor.pm
@@ -63,7 +63,7 @@ the file contents based on a new version of the reference file.
 =cut
 
 # FileEditor supports reading/editing a file
-sub _is_valid_source
+sub _is_valid_file
 {
     my ($self, $fn) = @_;
     return -f $fn;
@@ -78,9 +78,9 @@ sub _is_reference_newer
 {
     my ($self) = @_;
     my $is_newer = 0;   # Assume false
-    if (  exists(*$self->{options}->{source}) && $self->_is_valid_source(*$self->{options}->{source}) ) {
+    if (  exists(*$self->{options}->{source}) && $self->_is_valid_file(*$self->{options}->{source}) ) {
         # stat()[9] is modification time
-        if ( !$self->_is_valid_source(*$self->{filename}) || 
+        if ( !$self->_is_valid_file(*$self->{filename}) || 
              ((stat(*$self->{options}->{source}))[9] > (stat(*$self->{filename}))[9]) ) {
           $is_newer = 1
         }
@@ -107,7 +107,7 @@ sub new
     *$self->{options}->{source} = $opts{source} if exists ($opts{source});
     if ( $self->_is_reference_newer() ) {
         $src_file = *$self->{options}->{source};
-    } elsif ($self->_is_valid_source(*$self->{filename})) {
+    } elsif ($self->_is_valid_file(*$self->{filename})) {
         $src_file = *$self->{filename};
     }
     if ( $src_file ) {
--- a/src/main/perl/FileEditor.pm
+++ b/src/main/perl/FileEditor.pm
@@ -71,6 +71,7 @@ sub _is_reference_newer
     my ($self) = @_;
     my $is_newer = 0;   # Assume false
     if (  exists(*$self->{options}->{source}) && $self->_is_valid_source(*$self->{options}->{source}) ) {
+        # stat()[9] is modification time
         if ( !$self->_is_valid_source(*$self->{filename}) || 
              ((stat(*$self->{options}->{source}))[9] > (stat(*$self->{filename}))[9]) ) {
           $is_newer = 1
@@ -102,7 +103,7 @@ sub new
         $src_file = *$self->{filename};
     }
     if ( $src_file ) {
-        *$self->{LOG}->debug(2, "Reading initial contents from $src_file");
+        *$self->{LOG}->debug(2, "Reading initial contents from $src_file") if *$self->{LOG};
         my $txt = LC::File::file_contents ($src_file);
         $self->IO::String::open ($txt);
         $self->seek(IO_SEEK_END);
--- a/src/main/perl/FileEditor.pm
+++ b/src/main/perl/FileEditor.pm
@@ -47,15 +47,18 @@ the class constructor.
 =item new
 
 Returns a new object it accepts the same arguments as the constructor
-for C<CAF::FileWriter> with one additional option: C<source>. This
-option, when present, must be a file name whose contents will be used
+for C<CAF::FileWriter> with one additional option: 
+
+=over
+
+=item source 
+
+This option, when present, must be a file name whose contents will be used
 as the initial contents for the edited file if the source modification time
 is more recent than the edited file modification time. This allows to rebuild
-the file contents based on a new version of the reference/template file. 
+the file contents based on a new version of the reference file. 
 
-An example is:
-
-    my $fh = CAF::FileEditor('/my/file', source => '/ref/file');
+=back
 
 =cut
 
@@ -66,6 +69,11 @@ sub _is_valid_source
     return -f $fn;
 }
 
+# This method is only intended to be used in the context of the constructor
+# and entirely relies on the internal structure of the FileEditor object.
+# 'filename`  is the file name passed with instantiating the FileEditor and
+# 'options/sources' the 'source' parameter value (called a reference file
+# internally).
 sub _is_reference_newer
 {
     my ($self) = @_;
--- a/src/main/perl/FileEditor.pm
+++ b/src/main/perl/FileEditor.pm
@@ -78,11 +78,17 @@ sub _is_reference_newer
 {
     my ($self) = @_;
     my $is_newer = 0;   # Assume false
-    if (  exists(*$self->{options}->{source}) && $self->_is_valid_file(*$self->{options}->{source}) ) {
-        # stat()[9] is modification time
-        if ( !$self->_is_valid_file(*$self->{filename}) || 
-             ((stat(*$self->{options}->{source}))[9] > (stat(*$self->{filename}))[9]) ) {
+    if (  exists(*$self->{options}->{source}) ) {
+        # It is valid for the source value to be a pipe: in this case consider it
+        # as newer than an existing file.
+        if ( -p *$self->{options}->{source} ) {
           $is_newer = 1
+        } elsif ( $self->_is_valid_file(*$self->{options}->{source}) ) {
+        # stat()[9] is modification time
+            if ( !$self->_is_valid_file(*$self->{filename}) || 
+                 ((stat(*$self->{options}->{source}))[9] > (stat(*$self->{filename}))[9]) ) {
+               $is_newer = 1
+            }
         }
     }
     #FIXME: replace by $self->debug() after PR #154 has been merged...
--- a/src/main/perl/FileEditor.pm
+++ b/src/main/perl/FileEditor.pm
@@ -58,6 +58,9 @@ as the initial contents for the edited file if the source modification time
 is more recent than the edited file modification time. This allows to rebuild
 the file contents based on a new version of the reference file. 
 
+The C<source> can be a pipe: in this case, it is always considered more recent
+than the edited file.
+
 =back
 
 =cut
@@ -82,22 +85,21 @@ sub _is_reference_newer
         # It is valid for the source value to be a pipe: in this case consider it
         # as newer than an existing file.
         if ( -p *$self->{options}->{source} ) {
-          $is_newer = 1
+            $is_newer = 1
         } elsif ( $self->_is_valid_file(*$self->{options}->{source}) ) {
-        # stat()[9] is modification time
+            # stat()[9] is modification time
             if ( !$self->_is_valid_file(*$self->{filename}) || 
                  ((stat(*$self->{options}->{source}))[9] > (stat(*$self->{filename}))[9]) ) {
-               $is_newer = 1
+                $is_newer = 1
             }
         }
     }
     #FIXME: replace by $self->debug() after PR #154 has been merged...
     if ( *$self->{LOG} ) {
         if ( $is_newer ) {
-            # As this is a non reproducible event, be sure to log it when it happens
-            *$self->{LOG}->info("File ", *$self->{filename}, " contents reset to reference file (", *$self->{options}->{source}, ") contents.");
+            *$self->{LOG}->debug(1, "File ", *$self->{filename}, " older than reference file (", *$self->{options}->{source}, ",");
         } else {
-            *$self->{LOG}->debug(1, "Reference file (", *$self->{options}->{source}, ") older than ", *$self->{filename}, ": ignoring it");
+            *$self->{LOG}->debug(1, "Reference file (", *$self->{options}->{source}, ") older than ", *$self->{filename});
         };
     };
     return $is_newer;
@@ -112,6 +114,8 @@ sub new
 
     *$self->{options}->{source} = $opts{source} if exists ($opts{source});
     if ( $self->_is_reference_newer() ) {
+        # As this is a non reproducible event, be sure to log it when it happens
+        *$self->{LOG}->info("File ", *$self->{filename}, " contents reset to reference file (", *$self->{options}->{source}, ") contents.") if *$self->{LOG};
         $src_file = *$self->{options}->{source};
     } elsif ($self->_is_valid_file(*$self->{filename})) {
         $src_file = *$self->{filename};
--- a/src/main/perl/FileEditor.pm
+++ b/src/main/perl/FileEditor.pm
@@ -47,7 +47,15 @@ the class constructor.
 =item new
 
 Returns a new object it accepts the same arguments as the constructor
-for C<CAF::FileWriter>
+for C<CAF::FileWriter> with one additional option: C<source>. This
+option, when present, must be a file name whose contents will be used
+as the initial contents for the edited file if the source modification time
+is more recent than the edited file modification time. This allows to rebuild
+the file contents based on a new version of the reference/template file. 
+
+An example is:
+
+    my $fh = CAF::FileEditor('/my/file', source => '/ref/file');
 
 =cut
 
@@ -62,16 +70,20 @@ sub _is_reference_newer
 {
     my ($self) = @_;
     my $is_newer = 0;   # Assume false
-    if ( $self->_is_valid_source(*$self->{options}->{source}) ) {
+    if (  exists(*$self->{options}->{source}) && $self->_is_valid_source(*$self->{options}->{source}) ) {
         if ( !$self->_is_valid_source(*$self->{filename}) || 
              ((stat(*$self->{options}->{source}))[9] > (stat(*$self->{filename}))[9]) ) {
           $is_newer = 1
         }
     }
-    if ( $is_newer ) {
-        *$self->{LOG}->debug(2, "Source file (", *$self->{options}->{source}, " is newer than ", *$self->{filename}, ": use it");
-    } else {
-        *$self->{LOG}->debug(2, "Source file (", *$self->{options}->{source}, " older than ", *$self->{filename}, ": ignore it");
+    #FIXME: replace by $self->debug() after PR #154 has been merged...
+    if ( *$self->{LOG} ) {
+        if ( $is_newer ) {
+            # As this is a non reproducible event, be sure to log it when it happens
+            *$self->{LOG}->info("File ", *$self->{filename}, " contents reset to reference file (", *$self->{options}->{source}, ") contents.");
+        } else {
+            *$self->{LOG}->debug(1, "Reference file (", *$self->{options}->{source}, ") older than ", *$self->{filename}, ": ignoring it");
+        };
     };
     return $is_newer;
 }
@@ -80,16 +92,18 @@ sub new
 {
     my $class = shift;
     my $self = $class->SUPER::new (@_);
-    my $txt;
+    my $src_file;
     my ($path, %opts) = @_;
 
     *$self->{options}->{source} = $opts{source} if exists ($opts{source});
-    if ( exists(*$self->{options}->{source}) && $self->_is_reference_newer() ) {
-        $txt = LC::File::file_contents (*$self->{options}->{source});
+    if ( $self->_is_reference_newer() ) {
+        $src_file = *$self->{options}->{source};
     } elsif ($self->_is_valid_source(*$self->{filename})) {
-        $txt = LC::File::file_contents (*$self->{filename});
+        $src_file = *$self->{filename};
     }
-    if ( $txt ) {
+    if ( $src_file ) {
+        *$self->{LOG}->debug(2, "Reading initial contents from $src_file");
+        my $txt = LC::File::file_contents ($src_file);
         $self->IO::String::open ($txt);
         $self->seek(IO_SEEK_END);
     }
--- a/src/test/perl/test-caffileeditor_source.t
+++ b/src/test/perl/test-caffileeditor_source.t
@@ -5,11 +5,29 @@ use FindBin qw($Bin);
 use lib "$Bin/modules";
 use CAF::FileEditor;
 use Test::More tests => 7;
+use Test::MockModule;
 use Test::Quattor::Object;
 use Carp qw(confess);
 use File::Path;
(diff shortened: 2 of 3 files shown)


jouvin:
In the current version of the PR, the associated unit test `test-caffileeditor_source.t` doesn't work because `LC::File::file_contents` returns an `undef` value instead of the contents. I have no clue why this happens in the context of the test. Any help by the Perl experts appreciated!

src/main/perl/FileEditor.pm
```diff
@@ -58,13 +58,38 @@ sub _is_valid_source
...
+{
+    my ($self) = @_;
+    my $is_newer = 0;   # Assume false
+    if ( $self->_is_valid_source(*$self->{options}->{source}) ) {
+        if ( !$self->_is_valid_source(*$self->{filename}) || 
+             ((stat(*$self->{options}->{source}))[9] > (stat(*$self->{filename}))[9]) ) {
+          $is_newer = 1
+        }
+    }
+    if ( $is_newer ) {
```
stdweird: too much duplicated code, make this one line
```perl
*$self->{LOG}->debug(2, "Source file (", *$self->{options}->{source}, ") is ",($is_newer? "newer" : "older")," than ", *$self->{filename}, ": ",($is_newer ? "use" : "ignore")," it");
```

src/main/perl/FileEditor.pm
```diff
@@ -58,13 +58,38 @@ sub _is_valid_source
...
 sub new
 {
     my $class = shift;
     my $self = $class->SUPER::new (@_);
-    if ($self->_is_valid_source(*$self->{filename})) {
-        my $txt = LC::File::file_contents (*$self->{filename});
+    my $txt;
+    my ($path, %opts) = @_;
+
+    *$self->{options}->{source} = $opts{source} if exists ($opts{source});
```
stdweird: needs an entry in the pod as well

src/main/perl/FileEditor.pm
```diff
@@ -58,13 +58,38 @@ sub _is_valid_source
     return -f $fn;
 }
 
+sub _is_reference_newer
+{
+    my ($self) = @_;
+    my $is_newer = 0;   # Assume false
+    if ( $self->_is_valid_source(*$self->{options}->{source}) ) {
+        if ( !$self->_is_valid_source(*$self->{filename}) || 
+             ((stat(*$self->{options}->{source}))[9] > (stat(*$self->{filename}))[9]) ) {
```
stdweird: i'm not sure about `.rpm-new`, but rpms maintain timestamps of files, so files added by rpms at time X might have older timestamps.
also CAF (via LC) allows to set the mtime, so this is might give some unwanted sideeffects.
(but i like the idea of the PR)

src/main/perl/FileEditor.pm
```diff
@@ -58,13 +58,38 @@ sub _is_valid_source
...
 {
     my $class = shift;
     my $self = $class->SUPER::new (@_);
-    if ($self->_is_valid_source(*$self->{filename})) {
-        my $txt = LC::File::file_contents (*$self->{filename});
+    my $txt;
+    my ($path, %opts) = @_;
+
+    *$self->{options}->{source} = $opts{source} if exists ($opts{source});
+    if ( exists(*$self->{options}->{source}) && $self->_is_reference_newer() ) {
```
stdweird: the `exists()` is covered by the `_is_reference_newer`

src/main/perl/FileEditor.pm
stdweird: also needs `... if *$self->{LOG};` or the conditional logger from #154

src/main/perl/FileEditor.pm
```diff
@@ -58,13 +58,38 @@ sub _is_valid_source
...
+        $txt = LC::File::file_contents (*$self->{options}->{source});
```
stdweird: this needs to be logged at least with `verbose` that source is used

stdweird: also avoid code duplication. best to only modify the filename to read from, and use the `LC::File::file_contents`
even better, let `_is_reference_newer` return the correct filename to read from

src/main/perl/FileEditor.pm
```diff
@@ -58,13 +58,38 @@ sub _is_valid_source
...
+    } elsif ($self->_is_valid_source(*$self->{filename})) {
+        $txt = LC::File::file_contents (*$self->{filename});
+    }
+    if ( $txt ) {
```
stdweird: has to be `if (defined($txt))`

src/test/perl/test-caffileeditor_source.t
```diff
@@ -0,0 +1,66 @@
...
+use lib "$Bin/modules";
+use testapp;
+use CAF::FileEditor;
+use Test::More tests => 3;
+use Test::Quattor::Object;
+use Carp qw(confess);
+use File::Path;
+use File::Temp qw(tempfile);
+
+my $testdir = 'target/test/editor';
```
stdweird: the testdir has to be unique for the test, we cannot risk using directories with files from previous tests

src/test/perl/test-caffileeditor_source.t
```diff
@@ -0,0 +1,66 @@
...
+use constant TEXT => <<EOF;
+En un lugar de La Mancha, de cuyo nombre no quiero acordarme
+no ha tiempo que vivía un hidalgo de los de lanza en astillero...
+EOF
+use constant HEADTEXT => <<EOF;
+... adarga antigua, rocín flaco y galgo corredor.
+EOF
+
+my $fh;
+my $this_app = testapp->new ($0, qw (--verbose));
```
stdweird: remove the testapp code, you are not using it

src/test/perl/test-caffileeditor_source.t
```diff
@@ -0,0 +1,66 @@
...
+
+*testapp::error = sub {
+    my $self = shift;
+    $self->{ERROR} = @_;
+};
+
+# Create a file and check that it is empty
+($fh, $filename) = tempfile(DIR => $testdir);
+$fh->close();
+$fh = CAF::FileEditor->new($filename, log => $obj);
```
stdweird: use `CAF::FileReader` to read the contents of a file, this is confusing

src/test/perl/test-caffileeditor_source.t
```diff
@@ -0,0 +1,66 @@
...
+
+my $testdir = 'target/test/editor';
+mkpath($testdir);
+(undef, my $filename) = tempfile(DIR => $testdir);
+
+use constant TEXT => <<EOF;
+En un lugar de La Mancha, de cuyo nombre no quiero acordarme
+no ha tiempo que vivía un hidalgo de los de lanza en astillero...
+EOF
+use constant HEADTEXT => <<EOF;
```
stdweird: remove if not used

src/test/perl/test-caffileeditor_source.t
```diff
@@ -0,0 +1,66 @@
...
+is("$fh","","Existing file ($filename) empty");
+$fh->close();
+
+# Check that reference file contents is used as the initial contents when it
+# is newer than the file edited.
+sleep 5;
+my ($ref_fh, $ref_filename) = tempfile(DIR => $testdir);
+print $ref_fh TEXT;
+$ref_fh->close();
+$fh = CAF::FileEditor->new($filename, log => $obj, source => $ref_filename);
```
stdweird: use FileReader (looks odd, but is a subclass of Editor); but why do you cancel?

src/test/perl/test-caffileeditor_source.t
```diff
@@ -0,0 +1,66 @@
...
+($fh, $filename) = tempfile(DIR => $testdir);
+$fh->close();
+$fh = CAF::FileEditor->new($filename, log => $obj);
+$fh->cancel;
+is("$fh","","Existing file ($filename) empty");
+$fh->close();
+
+# Check that reference file contents is used as the initial contents when it
+# is newer than the file edited.
+sleep 5;
```
stdweird: set the mtime of file instead of sleeping

src/test/perl/test-caffileeditor_source.t
```diff
@@ -0,0 +1,66 @@
...
+print $ref_fh TEXT;
+$ref_fh->close();
+$fh = CAF::FileEditor->new($filename, log => $obj, source => $ref_filename);
+$fh->cancel;
+is("$fh",TEXT,"Reference file ($filename) contents used");
+$fh->close();
+
+# Check that reference file contents is not used as the initial contents when it
+# is older than the file edited.
+sleep 5;
```
stdweird: change the original file mtime, it's easier

src/test/perl/test-caffileeditor_source.t
```diff
@@ -0,0 +1,66 @@
...
+my ($new_fh, $new_filename) = tempfile(DIR => $testdir);
+$new_fh->close();
+$new_fh = CAF::FileEditor->new($new_filename, log => $obj, source => $ref_filename);
+$new_fh->cancel;
+is("$new_fh","","Existing file ($new_filename) contents used");
+$fh->close();
+
```
stdweird: test is missing
* existing of source attribute
* test of private method

src/main/perl/FileEditor.pm
stdweird: can you also add a comment what the 10th entry is in the stat result?

jouvin: 10th enty is the mod time and should not be affected by the creation time in the RPM.

src/test/perl/test-caffileeditor_source.t
jouvin: No reason in fact!

src/test/perl/test-caffileeditor_source.t
jouvin: The pb is that stat() as a 1s resolution... I plan to reduce it to s or 3s.

jouvin:
@stdweird thanks for all your comments, I'll implement them (as usual!). I was aware that the code was not polished. I concentrated on getting the test working and unfortunately I don't see anything in your comment that may help... Even the `defined($text)` is not the cause of the problem as `$txt` should be non empty in the context of the failing text. In fact is you print `$txt` after calling `LC::File::file_contents` (using `debug()` or friends), you will get a warning that `$txt` is undefined. No clue why as this is exactly the existing code that is working (but probably not really tested as it is generally mocked).

src/test/perl/test-caffileeditor_source.t
jouvin: I don't think it is a problem as the file name is generated (based on a time stamp I guess). And it makes the debugging easier IMO.

src/test/perl/test-caffileeditor_source.t
jouvin: I prefer to really test `FileEditor` as this is a test for it.

src/test/perl/test-caffileeditor_source.t
jouvin: In fact it is needed, at least the first one, as the file is reused in next step and we want it to be empty (else the test will succeed even if the code fails).

src/test/perl/test-caffileeditor_source.t
jouvin: I don't think that manipulating mtime is easier/cleaner.

src/test/perl/test-caffileeditor_source.t
jouvin: I tend to disagree as I first need to retrieve the existing one...

src/main/perl/FileEditor.pm
jouvin: Honnestly I don't buy this suggestion as a simplification...

jouvin:
@stdweird I did more tests to understand the problem of `LC::File::file_contents`. I have been able to reproduce the problem with the original `FileEditor` code (6b587a4e7b19234d1924b09fe30d6543e40245df). If you execute `test-caffileeditor_source.t` (after removing the 4 lines related to testing `source` and `_is_reference_newer()`) you obtain the same failing tests which is unexpected for the last test (file newer than reference). And if you add in the constructor after the call to `LC::File::file_contents` the following line:
```
        *$self->{LOG}->debug(2,*$self->{filename}." contents: $txt");
```

You get a warning that `$txt`  is uninitialized. As for me, the problem is related to the execution of `LC::File::file_contents` in the unit test context as this piece of code works well in the real life! No clue why...

jouvin:
Workaround for the failing `LC::File::file_contents` in the test context: mock it... Not glorious... but I didn't find a better solution as clearly (see previous comment) the problem is `LC::File::file_contents` itself, not the calling code.

src/main/perl/FileEditor.pm
```diff
@@ -58,13 +66,45 @@ sub _is_valid_source
     return -f $fn;
 }
 
+sub _is_reference_newer
```
ned21: This sub routine lacks any POD. It also hardcodes references to the internal data structures which makes it hard to follow: I have to know what *$self->{options}->{source} and *$self->{filename} refer to in order to grok it. Could it be made more general, so that the things/references to be compared are passed as arguments? Then the function could conceivably be reused in other places too.

jouvin: This is the usual stuff in CAF::Filexxx. There are already quite a lot of it around. I can add a comment that you should look at `CAF::FileWriter` for details. I can add a POD but this is a very internal method and as there were none for `_is_valid` I decided to do the same...

src/main/perl/FileEditor.pm
```diff
@@ -47,7 +47,15 @@ the class constructor.
 =item new
 
 Returns a new object it accepts the same arguments as the constructor
-for C<CAF::FileWriter>
+for C<CAF::FileWriter> with one additional option: C<source>. This
```
stdweird: make this extendable
```perl
with following addiotional options:
=over

=item source: ...

=back
```

src/main/perl/FileEditor.pm
```diff
@@ -47,7 +47,15 @@ the class constructor.
...
+option, when present, must be a file name whose contents will be used
+as the initial contents for the edited file if the source modification time
+is more recent than the edited file modification time. This allows to rebuild
+the file contents based on a new version of the reference/template file.
```
stdweird: remove the `/template`, nothing is templated here. i would also not use the term `reference` but stick with `source` to avoid confusion

src/main/perl/FileEditor.pm
```diff
@@ -47,7 +47,15 @@ the class constructor.
...
 
 Returns a new object it accepts the same arguments as the constructor
-for C<CAF::FileWriter>
+for C<CAF::FileWriter> with one additional option: C<source>. This
+option, when present, must be a file name whose contents will be used
+as the initial contents for the edited file if the source modification time
+is more recent than the edited file modification time. This allows to rebuild
+the file contents based on a new version of the reference/template file. 
+
+An example is:
```
stdweird: i'm not sure what the example adds here. it's an option like any other one

src/main/perl/FileEditor.pm
```diff
@@ -58,13 +66,45 @@ sub _is_valid_source
     return -f $fn;
 }
 
+sub _is_reference_newer
+{
+    my ($self) = @_;
+    my $is_newer = 0;   # Assume false
+    if (  exists(*$self->{options}->{source}) && $self->_is_valid_source(*$self->{options}->{source}) ) {
```
stdweird: ok, this is potentially dangerous, as `FileReader` inherits from Editor, but `valid_source` for reader can be a pipe, and the stat code below will not like that. so this PR needs modifcations for FileReader (unless you want editor to support pipes as `source`, which could make sense).

src/main/perl/FileEditor.pm
ned21: That was really my point - this is too much of an internal/single use specific function. If it was made more general and independent of the internal structures of this class then it could potentially be moved else where and used by other parts of the code base. It would also be easier to understand and test since the reader/test would only need to understand the function and not the surrounding state.

src/main/perl/FileEditor.pm
```diff
@@ -58,13 +66,45 @@ sub _is_valid_source
...
+        if ( !$self->_is_valid_source(*$self->{filename}) || 
+             ((stat(*$self->{options}->{source}))[9] > (stat(*$self->{filename}))[9]) ) {
+          $is_newer = 1
+        }
+    }
+    #FIXME: replace by $self->debug() after PR #154 has been merged...
+    if ( *$self->{LOG} ) {
+        if ( $is_newer ) {
+            # As this is a non reproducible event, be sure to log it when it happens
+            *$self->{LOG}->info("File ", *$self->{filename}, " contents reset to reference file (", *$self->{options}->{source}, ") contents.");
```
stdweird: both messages are not correct, as this method returns a simple boolean, it does not do anything (like resetting or ignoring).
i would just report the value of `is_newer` with debug 1 and report with info (nice catch on non-reproducible) later when the source is set.
ideally, the is_newer is a string with the reason why, something that can be added to the message, in this case eg. a timedifference or mtime of both files.

src/main/perl/FileEditor.pm
```diff
@@ -58,13 +66,45 @@ sub _is_valid_source
...
 {
     my $class = shift;
     my $self = $class->SUPER::new (@_);
-    if ($self->_is_valid_source(*$self->{filename})) {
-        my $txt = LC::File::file_contents (*$self->{filename});
+    my $src_file;
+    my ($path, %opts) = @_;
+
+    *$self->{options}->{source} = $opts{source} if exists ($opts{source});
+    if ( $self->_is_reference_newer() ) {
```
stdweird: the name of the method is and option are also a bity confusion: the source option is dealt with via the reference method, but the is also a `is_valid_source` method that has nothing to do with `source` option.

src/main/perl/FileEditor.pm
```diff
@@ -58,13 +66,45 @@ sub _is_valid_source
...
+        $src_file = *$self->{options}->{source};
+    } elsif ($self->_is_valid_source(*$self->{filename})) {
```
stdweird: this is differencet from the old code, where `filename` was always verified to be a valid filename, in this case, this check is skipped if source is used. if source is defined, filename should either no exist or be a file.
`CAF::Check` has methods for that, but if you don't want to use caf::Check, but the existence and directory tests in separate methods, as `-e` and `-d` can't be mocked.

src/test/perl/test-caffileeditor_source.t
```diff
@@ -0,0 +1,77 @@
+#!/usr/bin/perl
+use strict;
+use warnings;
+use FindBin qw($Bin);
+use lib "$Bin/modules";
```
stdweird: this is what loads dummy lc modules, remove it and you should be able to use `LC::File::file_contents` without the mocking

src/test/perl/test-caffileeditor_source.t
```diff
@@ -0,0 +1,77 @@
...
+    my $fname = shift;
+    my $fh;
+    open($fh, "<", $fname) || die ("failed to open $fname");
+    my $contents = join('', <$fh>);
+    return $contents;
+}
+$lcfile->mock("file_contents", \&read_file_contents);
+
+
+my $testdir = 'target/test/editor';
```
stdweird: unique test dirs per test, e.g. rename this to `editor_source`

src/test/perl/test-caffileeditor_source.t
```diff
@@ -0,0 +1,77 @@
...
+# Create a file and check that it is empty
+($fh, $filename) = tempfile(DIR => $testdir);
+$fh->close();
+$fh = CAF::FileEditor->new($filename, log => $obj);
+is("$fh","","Existing file ($filename) empty");
+$fh->close();
+
+# Check that reference file contents is used as the initial contents when it
+# is newer than the file edited.
+sleep 2;
```
stdweird: set the mtime of reference/source or filename using `LC::Check::status($fn, mtime => 123456)` instead of sleeping

src/test/perl/test-caffileeditor_source.t
```diff
@@ -0,0 +1,77 @@
...
+$ref_fh->close();
+$fh = CAF::FileEditor->new($filename, log => $obj, source => $ref_filename);
+is(*$fh->{options}->{source},$ref_filename,"File source is correctly defined");
+ok($fh->_is_reference_newer(),"Source file ($ref_filename) is newer than actual file ($filename)");
+is("$fh",$TEXT,"Reference file ($filename) contents used");
+$fh->close();
+
+# Check that reference file contents is not used as the initial contents when it
+# is older than the file edited.
+sleep 2;
```
stdweird: set the mtime as above

src/test/perl/test-caffileeditor_source.t
```diff
@@ -0,0 +1,77 @@
...
+my ($new_fh, $new_filename) = tempfile(DIR => $testdir);
+print $new_fh $ANOTHER_TEXT;
+$new_fh->close();
+$new_fh = CAF::FileEditor->new($new_filename, log => $obj, source => $ref_filename);
+is(*$new_fh->{options}->{source},$ref_filename,"File source is correctly defined");
```
stdweird: hmm, i didn't realise this. wouldn't it be better that the source option is deleted from the options if it is not used as source? that way we have a way to determine if the source was used as source or not. with the current code, presence of the source as option is non-conclusive.

src/main/perl/FileEditor.pm
jouvin: After thinking more about it, I prefer to keep it this way: an internal method valid only in the context of the constructor. I added comments (rather than POD) to make this explicit and explain the main attributes used.

src/main/perl/FileEditor.pm
jouvin: `/template` removed but I think that using `source` in the text will be more confusing as one could consider that the source is the filename you passed when instantiating the `FileEditor`. and `reference` as the option name may also be very strange/confusiong... I didn't manage to find a perfect solution! I mentioned in another comment that `source` parameter value is referred internally as reference.

src/main/perl/FileEditor.pm
jouvin: It was an attempt to help! Removed...

src/main/perl/FileEditor.pm
jouvin: How to easily distinguish a pipe from a standard file? If we agree that supporting pipes as `source` make sense, I propose that if it is a pipe, it is always considered newer than the existing file.

src/main/perl/FileEditor.pm
jouvin: I saw it, I'd suggest to rename `_is_valid_source()`!

src/main/perl/FileEditor.pm
jouvin: Nice catch! Clearly `filename` must be a valid filename in any case so that the changes can be written.

src/test/perl/test-caffileeditor_source.t
jouvin: I don't understand why... For me this is a much better test to do a real modification with a short time interval...

src/test/perl/test-caffileeditor_source.t
jouvin: I don't have a strong opinion. But I don't think we tend to delete options and this may make the troubleshooting more difficult... This is the role of the constructor to decide what to do. If it is really important to know if the source was used, I'd prefer to add an attribute rather than remove the option....

src/main/perl/FileEditor.pm
jouvin: @stdweird in fact I am not sure to completely understand your point. In the previous code, nothing prevented filename to be a directory for example. Just it was not read but when you were closing the file, you would have got an error. I don't think the new code is really different. But it may be good to add the test you suggest, i.e. be sure that the file wil be writable at the end. May be better to handle as a separate issue...

src/main/perl/FileEditor.pm
stdweird: uisng `-p`, see `_is_valid_source` in FileReader

src/main/perl/FileEditor.pm
stdweird: @jouvin heh, that is indeed the case. i was unaware of this. i opened https://github.com/quattor/CAF/issues/157, but i already do not see how this could be handled cleanly or what it would add :confused:

src/main/perl/FileEditor.pm
stdweird: also fine by me, do not forget to modify FileReader as well

src/test/perl/test-caffileeditor_source.t
stdweird: it adds 4 seconds to the tests, and setting the mtime by hand to some non-sensical value is much better way to control it than some random sleep value. it's as simple as
```perl
$time=time()
utime(undef, $time - 10, $filename); # make filename old enough
```

src/test/perl/test-caffileeditor_source.t
stdweird: ok, if the need ever arises, i guess we can add a source attribute to track actual source

src/test/perl/test-caffileeditor_source.t
jouvin: I don't see any dummy LC modules and removing this `use lib` doesn't help if I didn't do something wrong...

src/main/perl/FileEditor.pm
jouvin: I added pipe support for `source` but I am not sure how to unit test it... (btw, I don't think this is unit tested in FileReader too).

src/main/perl/FileEditor.pm
jouvin: Changed to `is_valid_file` but it seems it was not detected by GitHub.

src/test/perl/test-caffileeditor_source.t
stdweird: apologies again, you need code like https://github.com/quattor/CAF/blob/master/src/test/perl/kerberos-process.t#L4
the dummy LC modules are in the resources dir

src/main/perl/FileEditor.pm
```diff
@@ -47,24 +47,79 @@ the class constructor.
...
 
 Returns a new object it accepts the same arguments as the constructor
-for C<CAF::FileWriter>
+for C<CAF::FileWriter> with one additional option: 
+
+=over
+
+=item source 
+
+This option, when present, must be a file name whose contents will be used
```
stdweird: with current code, the source for Editor can also be a pipe (and maybe mention that if it is a pipe, it is always considered more recent)

src/main/perl/FileEditor.pm
```diff
@@ -47,24 +47,79 @@ the class constructor.
...
+# internally).
+sub _is_reference_newer
+{
+    my ($self) = @_;
+    my $is_newer = 0;   # Assume false
+    if (  exists(*$self->{options}->{source}) ) {
+        # It is valid for the source value to be a pipe: in this case consider it
+        # as newer than an existing file.
+        if ( -p *$self->{options}->{source} ) {
+          $is_newer = 1
```
stdweird: indentation is wrong?

src/main/perl/FileEditor.pm
```diff
@@ -47,24 +47,79 @@ the class constructor.
...
+        } elsif ( $self->_is_valid_file(*$self->{options}->{source}) ) {
+        # stat()[9] is modification time
```
stdweird: indentation is wrong

stdweird:
@jouvin some minor remarks left, LGTM otherwise

src/main/perl/FileEditor.pm
jouvin: Done.

jouvin:
Last comments hopefully addresssed.

stdweird:
@jouvin thanks a lot!

PR #164, opened 2016-05-18
Rule based editor: more unit tests + fixes

jouvin:
This PR adds unit tests that uncovered a few issues when using something else than a space as a keyword/value separator.  The new unit tests are a significant improvement in term of coverage of the editing methods.

This PR introduces no feature change and has no impact on the current user of the RBE (`ncm-xrootd` and `ncm-dpmlfc`)


jouvin:
Ready for merging.

stdweird:
LGTM

PR #166, opened 2016-05-31
Path: support move method

stdweird:
Add support for `move`ing (/renaming) files and directories

### DIFF
--- a/src/main/perl/Path.pm
+++ b/src/main/perl/Path.pm
@@ -7,16 +7,20 @@ use LC::Exception qw (throw_error);
 use Readonly;
 
 use File::Path qw(rmtree);
-use File::Copy qw(move);
 use File::Temp qw(tempdir);
 use File::Basename qw(dirname);
+
+# do not use qw(move) or
+# remove the qw() (move is in @EXPORT),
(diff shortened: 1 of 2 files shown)


jrha:
LGTM, could do with a brief description

stdweird:
@jrha is that sufficient?

src/main/perl/Path.pm
```diff
@@ -147,7 +153,7 @@ sub _reset_exception_fail
     if ($EC->error()) {
         # LC::Exception supports formatted stringification
         my $errmsg = ''.$EC->error();
-        $self->debug(1, "Ignoring/resetting previous existing error: $errmsg");
+        $self->debug(1, "Ignoring/resetting previous existing error$msg: $errmsg");
```
jrha: What does adding `$msg` here do? I've missed the context.

jrha:
Yep, thanks.

src/main/perl/Path.pm
stdweird: it's an additional argument you can pass, mainly to get an idea what method is resetting any previously existing fail or exception. very usefull while debugging.

PR #168, opened 2016-06-03
Process: factor out LC::Process calls for consistent behaviour

### LINKED ISSUE #167: CAF::Process, $? and NoAction
I noticed a regression in some ncm- components when running in --noaction mode after upgrading to 16.2. The component would report an ERROR that a command had failed, even though it was never run (this is NoAction mode). The ERROR was being logged as a result of a check on $?, which CAF::Process documents as the correct way to check for errors from sub-processes. Downgrading perl-CAF and perl-LC did not resolve the problem, it was only when I reverted both ncm-ncd and ccm that the problem went away. Known working version is 14.10, I was unable to test anything between 14.10 and 16.2.

In NoAction mode, CAF::Process never sets $? so my hypothesis is that something in ncm-ncd and ccm are invoking subprocesses that causes $? to have a stale value when checked by the component. 

Is it possible to override or reset the value of $? or would we need to execute /bin/true instead?

stdweird:
Also fixes #167

### DIFF
--- a/src/main/perl/Process.pm
+++ b/src/main/perl/Process.pm
@@ -1,21 +1,24 @@
-# ${license-info}
-# ${developer-info}
-# ${author-info}
-# ${build-info}
+#${PMpre} CAF::Process${PMpost}
 
-package CAF::Process;
+use parent qw(CAF::Object);
 
-use strict;
-use warnings;
 use LC::Exception qw (SUCCESS throw_error);
 use LC::Process;
-use CAF::Object;
+
 use File::Which;
 use File::Basename;
 
 use overload ('""' => 'stringify_command');
+use Readonly;
 
-our @ISA = qw (CAF::Object);
+Readonly::Hash my %LC_PROCESS_DISPATCH => {
+    output => \&LC::Process::output,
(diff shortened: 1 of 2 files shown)


ned21:
I need to look at the code more closely to understand how it works but I tested it on the machine that was failing and confirmed that it fixes #167.

I see this is tagged 16.6 but it's already June. Are we skipping 16.4?

stdweird:
@ned21 https://github.com/quattor/CAF/pull/168/files#diff-935c84ca8d8a9b56c500c070f0e885aaR152 is the fix for the reported issue, but instead of copy/pasting it 5 times, i refactored all common code. it now also has consistent reporting and we can very easily hook it to CAF::History

stdweird:
@ned21 wrt 16.4 @jrha was going to shift all milestones, so this will become 16.8. if you think this is buggy enough to tag it 16.4, i won't stop you :smile: (but i'd say it has been broken forever, this does not look like a regression). i'm unsure if the shift from 16.4 to 16.6 will allow a new PRs to be part of th next release. there are also still quite a few PRs open for 16.4 the need to merged.

PR #170, opened 2016-06-06
Path: disable chatty LC::Check by default unless NoAction

stdweird:
Similar to #98


stdweird:
@ned21 as this is for CAF::Path, there shouldn't be any legacy behaviour to delay this merge :smile:

PR #171, opened 2016-06-29
Path tests: use correct test messages

stdweird:
Does not fix anything, but avoids confusing test messages


src/test/perl/path.t
```diff
@@ -548,19 +548,15 @@ chmod(0000, $basetempdir);
 init_exception("temp directory creation failure NoAction=0 permission");
 
 ok(!defined($mc->directory($tempdir, temp => 1)),
-   "temp directory on broken symlink parent returns undef on failure tempdir");
+   "temp directory on parent w/o permissions returns undef on failure tempdir");
```
jrha: *without* or *write-others*?

stdweird:
@jrha clarified

Issue #146, opened 2016-03-26
CAF::Lock is it's own reporter


stdweird:
CAF::Lock probably does not log at all, ever, and even if it did, it doesn't follow the usual `log => $self` magic.
Need to figure out where CAF::Lock logging is initialised, and see if we can make if a `CAF::Object` that supports `log => $self` magic instead.

stdweird:
Was fixed in #173

PR #173, opened 2016-07-06
Lock: handle existing old-style locks

### LINKED ISSUE #172: Lock not compatible with old lock
With #132  merged, locks taken with the old code are not seen by the new code (the other way around is certainly not supported)

Old locks are files with a pid in them, and presence of the file indicates the lock is taken; nothing more.
To support old lockstyle locks, we only need to modify the method to test the locking: an old-style lock is present when the file exists and is not empty (old-style unlocking removes the file). not sure if we should handling stale old-style locks.

Showstopper for 16.6.0 release (https://github.com/quattor/release/issues/249).

stdweird:
Fixes #172 #146

### DIFF
--- a/src/main/perl/Lock.pm
+++ b/src/main/perl/Lock.pm
@@ -1,19 +1,17 @@
 #${PMpre} CAF::Lock${PMpost}
 
-use CAF::Object;
-use CAF::Reporter;
-
-use LC::Exception qw(SUCCESS throw_error);
+use CAF::Object qw(SUCCESS);
 use FileHandle;
 use File::stat; # overrides builtin stat
 use Fcntl qw(:flock);
 
-use Exporter;
 use vars qw(@ISA @EXPORT @EXPORT_OK);
 
-@ISA = qw(CAF::Reporter CAF::Object Exporter);
+# Only required to support legacy CAF::Reporter inheritance
+our @ISA;
+use parent qw(CAF::Object Exporter);
 
-@EXPORT_OK = qw(FORCE_NONE FORCE_ALWAYS FORCE_IF_STALE);
+our @EXPORT_OK = qw(FORCE_NONE FORCE_ALWAYS FORCE_IF_STALE);
 
 
 use constant FORCE_NONE     => 0;
@@ -32,7 +30,7 @@ CAF::Lock - Class for handling application instance locking
 
     use CAF::Lock;
 
-    $lock = CAF::Lock->new('/var/lock/quattor/spma');
+    $lock = CAF::Lock->new('/var/lock/quattor/spma', log => $reporter);
 
     unless ($lock->set_lock()) {...}
     unless ($lock->set_lock(10, 2) {...}
@@ -42,7 +40,7 @@ CAF::Lock - Class for handling application instance locking
 
 =head1 INHERITANCE
 
-    CAF::Reporter
+    CAF::Object
 
 =head1 DESCRIPTION
 
@@ -152,14 +150,33 @@ sub is_set
 
 Initialize the object.  Called by new(I<lockfilename>).
 
+Optional arguments
+
+=over
+
+=item C<log>
+
+A C<CAF::Reporter> object to log to.
+
+=back
+
 =cut
 
 sub _initialize
 {
-    my ($self, $lockfilename) = @_;
+    my ($self, $lockfilename, %opts) = @_;
 
     $self->{LOCK_SET} = undef;
     $self->{LOCK_FILE} = $lockfilename;
+
+    # This is intended to move away from some legacy code
+    # Do not copy this ever somewhere else
+    # Use: $self->{log} = $opts{log} if $opts{log};
+    if ($opts{log}) {
+        $self->{log} = $opts{log};
+    } else {
+        unshift(@ISA, 'CAF::Reporter');
+    }
 
     return SUCCESS;
 }
--- a/src/main/perl/Lock.pm
+++ b/src/main/perl/Lock.pm
@@ -5,6 +5,7 @@ use CAF::Reporter;
 
 use LC::Exception qw(SUCCESS throw_error);
 use FileHandle;
+use File::stat; # overrides builtin stat
 use Fcntl qw(:flock);
 
 use Exporter;
@@ -17,9 +18,9 @@ use vars qw(@ISA @EXPORT @EXPORT_OK);
 
 use constant FORCE_NONE     => 0;
 use constant FORCE_ALWAYS   => 1;
-use constant FORCE_IF_STALE => 2;  # for backwards compatibility only
-                                   # has no effect now
-
+# for backwards compatibility only
+# has no effect on newly created locks
+use constant FORCE_IF_STALE => 2;
 
 =pod
 
@@ -107,16 +108,21 @@ sub unlock
     my $self = shift;
     if ($self->{LOCK_SET}) {
         # if we forced the lock LOCK_FH can be undef
+
         if ($self->{LOCK_FH}) {
             unless (flock($self->{LOCK_FH}, LOCK_UN)) {
                 $self->error("cannot release lock: $self->{LOCK_FILE}");
                 return;
             }
-            $self->error("cannot close lock file: $self->{LOCK_FILE}")
-                unless $self->{LOCK_FH}->close();
+            # close the filehandle, clearing any previous content
+            unless ($self->{LOCK_FH}->close()) {
+                $self->error("cannot close lock file: $self->{LOCK_FILE}");
+            }
         }
+
         $self->{LOCK_SET} = undef;
         $self->{LOCK_FH} = undef;
+
     } else {
         $self->warn("lock not held by this application instance: $self->{LOCK_FILE}, nothing to unlock");
     }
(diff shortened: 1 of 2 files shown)


stdweird:
@jouvin should be able to test this by adding the rpm to the repo. i have no tested it myself.

src/main/perl/Lock.pm
```diff
@@ -193,6 +216,54 @@ sub _try_lock
     return SUCCESS;
 }
 
+
+# Before CAF PR#132, CAF::Lock used the presence of the lock file as
+# condition to determine if the lock was taken or not.
+# The file would hold the PID of the process that had the lock
+# (and unlock removed the file).
+# FORCE_IF_STALE would force
```
jouvin: I have the feeling that part of the sentence is missing... or that this is a leftover...

jouvin:
I tested the fix at LAL and I have the feeling that it works. But I encountered another problem with `ncm-spma` (https://github.com/quattor/configuration-modules-core/issues/801) and this is difficult to disentangle everything at this point... A new RC with this fix would to test everything as this is painful to add a new temporary RPM and that without doing it the upgrade first fails and this may also have some side effects (I don't think so).

ned21:
@ttyS4 - any comment on this?

stdweird:
@jouvin i cleaned up the code and comments, hope they are clearer now.
@ttyS4 can you carefully review the unlock code? the main idea is to handle the case of a new style instance that took the lock using an existing (stale) oldstyle file. i added cleanup code to make sure those are cleaned up once and for all.

jouvin:
LGTM... but definitely not an expert in CAF::Lock.

src/main/perl/Lock.pm
```diff
@@ -107,16 +107,38 @@ sub unlock
...
+
+        # REMOVE_LOCKFILE is set, use unlink to unlock
+        my $use_unlink = $self->{REMOVE_LOCKFILE} && -e $self->{LOCK_FILE};
+
         if ($self->{LOCK_FH}) {
-            unless (flock($self->{LOCK_FH}, LOCK_UN)) {
-                $self->error("cannot release lock: $self->{LOCK_FILE}");
-                return;
+            if (! $use_unlink) {
+                # unlock will be done by unlink. between releasing the lock here
```
ned21: I don't follow this comment. The code below will only be run unlink if $use_unlink is true?

ned21:
I am struggling to determine whether this code is "correct" because I don't understand what it's intending to do. Ignoring the code for a moment, here's what I think a process could look like:

1. Take new-style lock. If succeed then guaranteed no other new-style processes can run.
2. Take old-style lock. This guarantees no other old-style processes can run.
3. If (2) fails then drop new-style lock.

i.e. wrap the old-style lock in new-style. I think this is the only way to guarantee safe upgrades from any older version of CAF::Lock.  But that means we have to keep old-style lock code around forever.

Another option would be:

1. Take new-style lock.
2. Check for presence of old-style lock. If present then, check for stale-ness and if not stale then abort. If stale, then remove it and carry on. 

This is I think closer to what you intend based on the comment in #172. It works because on a working system once the old-style process that has upgraded CAF::Lock has run has finished then the old lock files should never be used again. The check for stale-ness could even be dropped: if the file exists then assume the upgrade is still running and abort.

src/main/perl/Lock.pm
```diff
@@ -107,16 +107,38 @@ sub unlock
...
+                # and unlinking below, another process to get the lock
```
jouvin: I think a word is missing in the comment...

src/main/perl/Lock.pm
```diff
@@ -173,18 +195,35 @@ sub _try_lock
...
-    unless ($lf) {
-        $self->error("cannot create lock file: $self->{LOCK_FILE}");
-        return;
+    my $lf;
+
+    # has_lock: does the current instance have the lock?
+    my $has_lock = 1;
+
+    if ($self->_is_locked_oldstyle($force)) {
+        # Do not bother try to take the lock
```
jouvin: Comment unclear for me...

src/main/perl/Lock.pm
```diff
@@ -193,6 +232,57 @@ sub _try_lock
...
 
+
+# Before CAF PR#132, CAF::Lock used the presence of the lock file as
+# condition to determine if the lock was taken or not.
+# The file would hold the PID of the process that had the lock
+# (and unlock simply removed the file).
+# This method provides support for the existence of an old-style lockfile
+# held by another process using old-style locking code.
+# It does not modify/cleanup the old-style file (even if it is stale).
+# Returns boolean true is the lock taken by another process (i.e. if
```
jouvin: `Returns true if the lock has been taken...`

jouvin:
I read again the code after @ned21 [comment](https://github.com/quattor/CAF/pull/173#issuecomment-230901523) and I have the feeling that what is done is correct but that comments should really be improved to better document what is done if an old-style lock is found.

I have the feeling that the difference between what @stdweird implemented and @ned21 suggested (alternative) approach is that the old-style lock, if present, is not flocked (new-style lock). But I am not sure there is any practical difference as CAF::Lock is the only user of this lock and any new-style instance of CAF::Lock will properly honour the old-style lock as long as it is present.

stdweird:
@ned21 @jouvin my hasty attempt at a fix is not correct, working on new code, with tests this time; and hopefully without the unlink code.

stdweird:
I also fixed #146 so i had debugging output while adding tests, my apologies for that

@ned21 no more unlink

it should now handle existing old-style locks, but it does not play nice with old style locks; in particular, an old style lock cannot be taken anymore (unless FORCE is used, but that code clearly doesn't care that much)

if you want features/changes, please suggest a unittest so i can test things and then try to figure out how to make it work

jouvin:
I have not looked at the new code yet but tried it both for upgrade from and downgrade to a previous version. I am not completely sure that the downgrade scenario works, i.e. that the old `CAF::Lock` see the new-style lock (file should be the same). This is probably not very important once the upgrade has been done successfully but it may be important if something goes wrong and you need to downgrade. According to my tests (https://github.com/quattor/release/issues/249#issuecomment-231052241), downgrade seems to work properly. To be confirmed with RC2.

src/test/perl/lock.t
```diff
@@ -38,4 +46,90 @@ ok($lock2->set_lock(), "lock2 set when lock1 unlocked");
...
+   "Presence of non-empty lockfile is not an old-style lock (by other process) if FORCE_ALWAYS is used and pid exists");
+ok($lock1->_is_locked_oldstyle(FORCE_IF_STALE),
+   "Presence of non-empty lockfile is an old-style lock if FORCE_IF_STALE is used and pid exists");
+
+# Make oldstyle stale lock
+$lock1 = undef;
+unlink(LOCK_TEST);
+
+# will be a while till this is becomes a possible PID;
+# should also not bee too  big so perl still thinks this is an interger
```
jouvin: bee -> be
so -> so that ???

src/test/perl/lock.t
```diff
@@ -38,4 +46,90 @@ ok($lock2->set_lock(), "lock2 set when lock1 unlocked");
...
+
+sub readfile
+{
+    open(FH, shift);
+    my $txt = join('', <FH>);
+    close(FH);
+    return $txt;
+}
+
+# Make oldstyle non-stale lock
```
jouvin: oldstyle -> an old-style

src/test/perl/lock.t
```diff
@@ -38,4 +46,90 @@ ok($lock2->set_lock(), "lock2 set when lock1 unlocked");
...
+ok(! $lock1->_is_locked_oldstyle(FORCE_ALWAYS),
+   "Presence of non-empty lockfile is not an old-style lock (by other process) if FORCE_ALWAYS is used and pid exists");
+ok($lock1->_is_locked_oldstyle(FORCE_IF_STALE),
+   "Presence of non-empty lockfile is an old-style lock if FORCE_IF_STALE is used and pid exists");
+
+# Make oldstyle stale lock
+$lock1 = undef;
+unlink(LOCK_TEST);
+
+# will be a while till this is becomes a possible PID;
```
jouvin: will be -> will take
this is becomes -> this becomes

src/test/perl/lock.t
```diff
@@ -38,4 +46,90 @@ ok($lock2->set_lock(), "lock2 set when lock1 unlocked");
...
+
+$lock1 = CAF::Lock->new(LOCK_TEST, log => $obj);
+ok($lock1->_is_locked_oldstyle(),
+   "Presence of non-empty lockfile is an old-style lock (by other process) and stale pid");
+ok(! $lock1->_is_locked_oldstyle(FORCE_ALWAYS),
+   "Presence of non-empty lockfile is not an old-style lock (by other process) if FORCE_ALWAYS is used and stale pid");
+ok(! $lock1->_is_locked_oldstyle(FORCE_IF_STALE),
+   "Presence of non-empty lockfile is not an old-style lock if FORCE_IF_STALE is used and stale pid");
+
+# Take a FORCE_IF_STALE lock
```
jouvin: I don't have the feeling that `FORCE_IF_STALE` is a type of lock... what aobut "Test FORCE_IF_STALE flag used to take a lock when a an old-style lock file is present but the pid which created it no longer exists"

src/main/perl/Lock.pm
```diff
@@ -107,16 +105,20 @@ sub unlock
...
         # if we forced the lock LOCK_FH can be undef
+
         if ($self->{LOCK_FH}) {
             unless (flock($self->{LOCK_FH}, LOCK_UN)) {
                 $self->error("cannot release lock: $self->{LOCK_FILE}");
                 return;
             }
+            # close the filehandle, clearing any previous content
             $self->error("cannot close lock file: $self->{LOCK_FILE}")
                 unless $self->{LOCK_FH}->close();
```
jouvin: merging on one line would make it easier to read (or using an unless block): I had the feeling at the first look that the error was unconditionally raised....

stdweird:
@jouvin thanks for the feedback, i squashed the fixes

jouvin:
Thanks for the fix! LGTM. Waiting for @ned21 feedback if he is available... As for me can be merged at any time to make an RC2...

src/main/perl/Lock.pm
```diff
@@ -193,6 +235,61 @@ sub _try_lock
...
+# (i.e. if it is locked by another old-style instance).
+sub _is_locked_oldstyle
+{
+    my ($self, $force) = @_;
+
+    my $other_has_lock = 0;
+
+    # No file -> no lock
+    # Empty file (like a lock with current code) -> no old-style lock
+    my @st = stat($self->{LOCK_FILE});
```
ned21: Why use stat here rather than just -e ? The long if() statement requires a lot of reading to understand nothing happens. I think it would be better to "short circuit" this sub in the routine case:
```
# use stat to detect file existence because ...
my @st = stat($self->{LOCK_FILE});
if( ! @st && $st[7] > 0 ) {
  # No file -> no lock
  return 0;
} else {
  my $other_has_lock = 0;
  <<current code>>
}
```

src/main/perl/Lock.pm
ned21: How about this wording:
```# Old style lock detected, cannot continue```

File: src/main/perl/Lock.pm (hunk 11 above)
Comment (stdweird) [inline src/main/perl/Lock.pm]: there are 2 conditions that immediately rule out an old style lock:
* no file
* empty file (like a new-style lock, or the leftover of a newstyle released lock)

hence the single stat to determine both

**Comment by stdweird** [comment]
@ned21 i removed the leftover confusing comment about the file removal.

**Comment by ttyS4** [comment]
I am away and will be able to check by the weekend. 

Regards, 
Tamas

File: src/main/perl/Lock.pm (hunk 11 above)
Comment (ned21) [inline src/main/perl/Lock.pm]: OK, that makes sense. Can we use the named attributes of stat instead of the numeric one? That would make the code more readable without having to grub around in man pages to remember what the $st[7] represents.

My other point was purely stylistic. The "common case" is that we are in a new-lock only situation. Therefore I should be able to read this check, see that it does nothing and ignore the rest of the code. 

```
if( rare-condition )  {
   # ugly stuff
}
return;
```
is fine until <<ugly stuff>> gets so long and deep that it's hard to figure out exactly where I return to the main code branch. In which case I think convention is to "short cut" the common case:
```
if (!rare-condition) {
  return;
} else {
  # long and deep code block
}
```
I think this is something recommended in the _Code Complete_ book but I don't remember anymore.

Thinking a bit more, an even better approach would be to move the stat() up to the calling point and hide the ugly stuff entirely in the subroutine.  

```
my @st = stat($self->{LOCK_FILE});
if (@st && $st->size > 0 && $self->_is_locked_oldstyle($force)) {
```

stdweird:
@ned21 i cleaned up the logic a bit, but i only return at the end of the function (maybe you can guess who made enough comments about that in all my other PRs :smile: )
i tought about moving the size and existence check to `_try_lock`, but they really have nothing to do with newstyle locking, so i kept it out.

jrha:
AFAICT this is fine now, any objections to merging it so I can build RC2?

stdweird:
@ttyS4 even if this is already merged, i'd really appreciate your feedback on this PR

ttyS4:
@stdweird In general I am ok with it. There are a few minor things that I would change however:

`if ! defined` these should be `unless defined` - yeah i know :-)
`$has_lock` we should not initialize this variable

force_stale - this should be the default behavior (in the legacy detection code). No one really uses or asks for it so if we do not make it default it wont be invoked. Because it is compatibility code it is also ok to leave as it is. (force_stale was one of the first things I removed from the new locking logic.)

I was thinking about if `_is_locked_oldstyle` should care about force at all (the following makes sense after the previous suggestion). There are reason for and also against it. The new locking logic just sets the lock internally if flock fails *and* forced without touching the file like: "ok whatever, you have the lock".

Currently `_is_locked_oldstyle` is actually more like `_is_locked_oldstyle_or_did_it_told_you_to_lie_to_me` ;-)

Oh, it is late now. I hope it makes sense.

stdweird:
@ttyS4 thanks for the feedback. 
i agree that `force_stale` *should* have been the default, but there's no point in changing the legacy default behaviour (and that's why i kept the same force logic in the `is_locked_oldstyle`). and the new-style doesn't suffer from this anyway.
the oldstyle locking code is a way to handle existing/in-use oldstyle locks. as soon as a new style lock is taken, there's no turning back (and that's a good for locking; but not so much for backwards compatibility, like during a rollback).

jouvin:
About rollback, note that as said in [16.6.0-rc3](https://github.com/quattor/release/issues/250#issuecomment-231746432) issue tracker, the Quattor client downgrade works despite the first time the old `CAF::Lock` runs it doesn't see the newstyle lock. As for me, this is enough: there is no reason to support a wider use of a mix of oldstyle and newstyle locking.

jrha:
Thanks @jouvin.

PR #175, opened 2016-07-11
Lock: set_lock correct number of retries (and test)

stdweird:
Depending on the interpretation of the pod of `set_lock`, one retry was missing.
(Also adds improved error message and tests)


jrha:
LGTM

PR #176, opened 2016-07-14
Object: add conditional log method

stdweird:
Add `log` method to list of conditional log methods, so all log methods from `Reporter` now have a conditonal couterpart (needed for https://github.com/quattor/ncm-ncd/issues/70)


jrha:
Can you add a description please? Explain what this is for and what will use it.

stdweird:
@jrha better?

jrha:
Much!

PR #178, opened 2016-07-26
Reporter: handle non-printable characters

### LINKED ISSUE #177: Kerberos: binary garbage in debug message during init
```
[DEBUG] _gssapi_init status GSSAPI::Context::init->() args GSSAPI::Context=SCALAR(0x2ca1510), <undef>, GSSAPI::Name=SCALAR(0x27d9960), { 1 2 840 113554 1 2 2 }, 0, 0, <undef>, , { 1 2 840 113554 1 2 2 }, `�~	*�H��n�i�����na�j0�f��
                                                                                                              MYREALM�$0"��0hosthost.example.com��'�#�����I�����|�8��
                                                       8��r�)�ׁ���
E����J_�=�k3ĸ�l���E �O���j�ዦ>O��(��!�Z
��9f�q����;���㽑r�������ʵK5�C�q ����?0�˿o�fS
4�!��J5�� :.����0�ޠ�����
             �'��9���O� �����#q�M@nW�IRħ�ׄ�����Gs>�vR��qN�Ō'���)�(�b.�����Xߣ�F3��3٭
                                                                                                                           L�b��7M������K^I
��u��, 304, 86399 eval ec 
[VERB] GSS Success GSSAPI::Context::init: MAJOR: The routine completed successfully MINOR: Unknown error
``

### PR DESCRIPTION
**Pull request by stdweird**
Fixes #177

### DIFF
--- a/src/main/perl/Reporter.pm
+++ b/src/main/perl/Reporter.pm
@@ -45,6 +45,27 @@ sub _rep_setup
     return $_REP_SETUP;
 }
 
+# Join all arguments into a single string
+# Handles undefined arguments and non-printable garbage
+sub _make_message_string
+{
+    # Ensure that there is no undefined arg: replace by <undef> if any, force stringification otherwise.
+    my @args = map {defined($_) ? "$_" : '<undef>'} @_;
+
+    # Make single string
+    my $string = join('', @args);
+
+    # Replace any non-printable character with a ?
+    # print charclass only includes whitespace that is not a control charater
+    #    (so also add the space charclass (for e.g. newline))
+    # TODO: also exclude the cntrl charclass?
+    # TODO: this will remove non-ascii characters, incl. unicode;
+    #       unless 'use utf8' (and similar) is used. :print: means printable in current encoding
+    $string =~ s/[^[:print:][:space:]]/?/g;
+
+    return $string;
+}
+
 =pod
 
 =head1 NAME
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: src/main/perl/Reporter.pm [hunk 1]
```diff
@@ -45,6 +45,25 @@ sub _rep_setup
...
+    # Ensure that there is no undefined arg: replace by <undef> if any, force stringification otherwise.
+    my @args = map {defined($_) ? "$_" : '<undef>'} @_;
+
+    # Make single string
+    my $string = join('', @args);
+
+    # Replace any non-printable character with a ?
+    # negated character classes in this form are a perlism (perldoc perlrecharclass)
+    # TODO: this will remove non-ascii characters, incl. unicode;
+    #       unless 'use utf8' (and similar) is used. :print: means printable in current encoding
```
jouvin: I am not sure to understand... Does it mean that this is the responsibility of the caller to execute `use uf8`. Should not it be the default?

stdweird: yes, the caller should always take care of proper encoding. 
`use utf8` is a bit a an odd example, as it is only required if the perl code itself is in utf8, like the example in the unittest. so it should not be enabled by default.

stdweird:
@jrha thanks for merging, but this can use some in the field testing with non-standard locale
@jouvin @ned21 you have some machines lying around that have utf8 in log messages?

PR #179, opened 2016-07-31
Kerberos: support credential request and check

stdweird:


src/main/perl/Kerberos.pm
```diff
@@ -314,6 +319,18 @@ sub get_context
...
 
+    my $cred;
+
+    if ($opts{usecred}) {
+        $self->verbose("trying to get credential");
+        $cred = $self->get_cred(name => $name);
+        # Already reports error
+        return if ! $cred;
+    } else {
+        $self->verbose("No checking for credential");
```
ned21: This adds a message to every invocation if not using the new option. Should this be debug or is it important enough to be a warning? Also, "No" -> "Not" ?

stdweird: i'm not sure. the context mechanism is used to get the initial tokens, but without passing a credential instance, the principal passed via the name instance is used in a sort of best-effort way.

ned21: OK, let's leave it as verbose then. Can the messages be made consistent?
"trying to get credential" -> "Trying to get credential".
"No checking for credential" -> "Not checking for credential"

stdweird:
@ned21 addressed remark

PR #180, opened 2016-08-05
FileWriter: support for setting file modification time via mtime option

stdweird:
Fixes #169

PR #181, opened 2016-08-05
FileWriter: add support for keeps_state to override the global NoAction on per-file basis

stdweird:
This is useful for e.g. writing temporary files.

Depends on #180 

Interface is now similar to `Path` and `Process`

### DIFF
--- a/src/main/perl/FileWriter.pm
+++ b/src/main/perl/FileWriter.pm
@@ -117,6 +117,16 @@ File's modification time.
 
 Path for the backup file, if this one has to be re-written.
 
+=item keeps_state
+
+A boolean specifying whether a file change respects the current system
+state or not. A file with C<keeps_state> will be created/modified,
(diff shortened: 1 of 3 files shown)


stdweird:
@jrha feel free to change the milestone to 16.10

src/main/perl/FileWriter.pm
```diff
@@ -117,6 +117,15 @@ File's modification time.
 
 Path for the backup file, if this one has to be re-written.
 
+=item keeps_state
+
+A boolean specifying whether a file change respects the current system
+state or not. A file with C<keeps_state> will be created/modified,
+regardless of any value for C<NoAction>.
```
ned21: Add: "This is useful when creating temporary files that are required for a NoAction run."

stdweird:
@ned21 remark fixed

PR #202, opened 2016-11-26
FileWriter: use atomic write

### LINKED ISSUE #148: CAF::FileWriter updates files in-place
It should create a new file and rename to do it atomically.

### LINKED ISSUE #189: FileWriter: add diff to history

### LINKED ISSUE #223: CAF::FileWriter Can't call method "can" on unblessed reference at line 353
As reported in https://github.com/quattor/configuration-modules-core/pull/1036#issuecomment-283085607, getting the following error in a Jenkins run:

```
Can't call method "can" on unblessed reference at /var/lib/jenkins/perl/lib/perl5/CAF/FileWriter.pm line 353.
# Looks like your test exited with 255 before it could output anything.
src/test/perl/ips-configure.t ............. 
Failed 12/12 subtests 
```

stdweird:
Requires #204 and #206
Fixes #148
Fixes #223
Also fixes #189 

`File::AtomicWrite` requires 0.18, rpm can be build using `cpanspec -b --packager quattor File::AtomicWrite --force`


stdweird:
This requires testing obviously

src/main/perl/FileWriter.pm
```diff
@@ -179,43 +180,51 @@ secure way (not following symlinks, etc).
...
+
+        my $msg = 'was';
+
+        if ($changed) {
+            if($self->is_verbose()) {
                 $self->verbose ("Changes to $filename:");
                 $self->report ($diff);
+            }
+
+            if (*$self->{options}->{noaction}) {
```
ned21: Is this test the right way round? i.e. this seems to be the block that's executed when noaction is true?

src/main/perl/FileWriter.pm
```diff
@@ -210,15 +246,33 @@ sub close
...
+                };
+                # No need to check if owner exists. :groupname is supported
+                $opts->{owner} .= ":$options->{group}" if exists($options->{group});
+
+                eval {
+                    File::AtomicWrite->write_file($opts);
+                    $modified = 1;
+                };
+                if ($@) {
+                    $self->verbose("AtomicWrite gave error: $@");
```
ned21: Is this only verbose because throw_error takes care of generating the actual error? In which case debug would be more appropriate to avoid the duplicate message?

src/test/resources/File/AtomicWrite.pm
```diff
@@ -0,0 +1,79 @@
+package File::AtomicWrite;
+
+use strict;
+use warnings;
+
+use version;
+
+# Always recent enough
+our $VERSION = 10000.0.0;
```
ned21: I think we should declare here the API version we are mocking. Then if a later version changes, we get a failure and get reminded to go and check we updated the mock to match the behaviour of the real thing?

src/test/resources/File/AtomicWrite.pm
```diff
@@ -0,0 +1,79 @@
...
+
+=pod
+
+=head1 SYNOPSIS
+
+This is a backup module for File::AtomicWrite.
```
ned21: s/backup/mock/ ?

src/test/perl/modules/filetools.pm
```diff
@@ -0,0 +1,31 @@
...
+
+our @EXPORT_OK = qw(readfile makefile);
+
+# cannot use mocked filewriter
+sub makefile
+{
+    my $fn = shift;
+    my $dir = dirname($fn);
+    mkpath $dir if ! -d $dir;
+    open(FH, ">$fn");
```
ned21: Bare filehandles and two element open is considered dangerous. Please use ```open(my $fh, ">", "$fn");``` as per the Perl Best Practice Book.

src/test/perl/modules/filetools.pm
```diff
@@ -0,0 +1,31 @@
...
+    print FH (shift || "ok");
+    close(FH);
+}
+
+sub readfile
+{
+    open(FH, shift);
```
ned21: ditto

src/test/perl/modules/filetools.pm
```diff
@@ -0,0 +1,31 @@
...
+
+# cannot use mocked filewriter
+sub makefile
+{
+    my $fn = shift;
+    my $dir = dirname($fn);
+    mkpath $dir if ! -d $dir;
+    open(FH, ">$fn");
+    print FH (shift || "ok");
+    close(FH);
```
ned21: Neither open() nor close() do any error handling which is generally bad, although given this is a unit test helper function, perhaps it's not bad in this specific case?

src/main/perl/FileWriter.pm
stdweird: probably. but this commit is already outdated

src/test/resources/File/AtomicWrite.pm
stdweird: done

src/test/resources/File/AtomicWrite.pm
stdweird: done

src/test/perl/modules/filetools.pm
stdweird: done. now `die`s with proper message

stdweird:
@ned21 first round of remarks addressed :smile:

src/test/perl/modules/filetools.pm
```diff
@@ -0,0 +1,33 @@
+package filetools;
```
stdweird: switch to new unittest Filetools module

stdweird: @jrha @ned21 hmm, i can't request changes on my own PR. can you help me not to forget 😄

jrha: Don't forget to switch to new unittest Filetools module.

stdweird:
@jrha Filetools used
@ned21 what changes do you still need?

jouvin:
I'd like to test it to see in particular if it fixes https://github.com/quattor/maven-tools/issues/139. Could we get the dependencies in quattor externals repo?

stdweird:
@jouvin i mailed them to you and @jrha lasty week 😄

stdweird:
@ned21 i added another commit to fix #223

ned21:
The fix to #223  is 2fee451? Can the "Fixes #223" line be added to the commit log? Also any chance of a separate PR in case this doesn't make the cut for 17.3? Then we can get the spma change merged.

stdweird:
@ned21 this is not blocking the spma PR at all, the fix is just to handle the failure in cleaner way, it doesn 't fix the failure itself

stdweird:
@ned21 message updated

ned21:
I'm good with this PR, waiting on @jrha  to confirm the changes he requested have been made.

stdweird:
rpms added to externals repo

PR #204, opened 2016-12-12
Reporter: add structured logging and message templating

stdweird:
Requires #206


jrha:
Does this need new build tools or anything else poking to pass tests?

stdweird:
no, i just rebased

PR #209, opened 2017-01-13
Handle Download::LWP on el5/el6

stdweird:


stdweird:
@jrha @ned21 this is required to let the unittests pass on el5
however, this also exposes a potential serious backwards incompatibility issue: `verify_hostname` is enforced on EL6 and EL7; meaning that ill-configured systems (eg no access to CA cert of https server) will now fail when using this (eg now that https://github.com/quattor/CCM/pull/156 is merged, ccm-fetch will fail for such setups).

stdweird:
seems also required on el6.

PR #221, opened 2017-02-28
Download/LWP and Kerberos: disable unitialized warnings while making local ENV copy

stdweird:
ncm-download unittests fail on el5 due to these warnings


jrha:
Removing support may well be some time in the future...

PR #224, opened 2017-03-03
CAF::Application: add methods option_exists() and set_options()

### LINKED ISSUE #210: Application.pm: no possibility to test if an option exists
If I am right, there is currently no way to test if an option is defined in the context of an application. If using `option()` method with a non existing option, an error is printed saying that the option doesn't exist. It would be great to have either `option()` testing the exististence of the option and silently ignoring non existing option (this will result in an `undef` value returned).or to have a method to test the existence of the option. 

My preference would be to have `option()` ignore silently missing variables with a message logged if debug level >= 1. Testing a variable existence is as simple as:
```
$self->{CONFIG}->_exists($opt)
```

Having it done by `option()` would not require any change in existing code to get rid of the error message that happens if a variable is not defined. When a variable is required to exist, it will be enough to test if the value returned by `option()` is `undef`, as it is already the case.

jouvin:
Both methods are wrapper of `AppConfig` methods to avoid exposing `CAF::Application` and `AppConfig` implementation details
- option_exists(): allow to check the existence of a given option
  - Used in option() to avoid the AppConfig error if the option doesn't exist
- set_option(): allow to define an option and its value

Fixes #210

### DIFF
--- a/src/main/perl/Application.pm
+++ b/src/main/perl/Application.pm
@@ -125,7 +125,7 @@ sub username
 
 Returns true if the option exists, false otherwhise. Option can be
 defined either in the application configuration file or on the
-command line (based on AppConfig module).
+command line (based on C<AppConfig> module).
 
 =cut
 
@@ -160,6 +160,28 @@ sub option
     }
 
     return $default;
+}
+
+=pod
+
+=item set_option($opt, $val): SUCCESS
+
+Defines an option and sets its value. If the option was previously
+defined, its value is overwritten. This is a wrapper over C<AppConfig>
+methods to hide the internal implementation of a C<CAF::Application>.
+
+This method always returns SUCCESS.
+
+=cut
+
+sub set_option
+{
+    my ($self, $opt, $val) = @_;
+
+    $self->{'CONFIG'}->define($opt);
+    $self->{'CONFIG'}->set($opt, $val);
+
+    return SUCCESS;
 }
 
 =pod
--- a/src/main/perl/Application.pm
+++ b/src/main/perl/Application.pm
@@ -121,20 +121,45 @@ sub username
 
 =pod
 
+=item option_exists($opt): boolean
+
+Returns true if the option exists, false otherwhise. Option can be
+defined either in the application configuration file or on the
+command line (based on AppConfig module).
+
+=cut
+
+# Check if a configuration option exists
+sub option_exists
+{
+    my ($self, $option) = @_;
+    return $self->{CONFIG}->_exists($option);
+}
+
+=pod
+
 =item option($opt): scalar|undef
 
 Returns the option value coming from the command line and/or
 configuration file. Scalar can be a string, or a reference to a hash
 or an array containing the option's value. option() is a wrapper
-on top of AppConfig->get($opt).
+on top of AppConfig->get($opt). 
+
+If the option doesn't exist, returns C<undef>, except if the C<default>
+argument has been specified: in this case this value is returned but
+the option remains undefined.
 
 =cut
 
 sub option
 {
-    my ($self,$opt) = @_;
+    my ($self, $opt, $default) = @_;
 
-    return $self->{'CONFIG'}->get($opt);
+    if ( $self->option_exists($opt) ) {
+        return $self->{'CONFIG'}->get($opt);
+    }
+
+    return $default;
 }
 
 =pod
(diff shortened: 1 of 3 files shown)


src/main/perl/Application.pm
```diff
@@ -121,20 +121,62 @@ sub username
...
 or an array containing the option's value. option() is a wrapper
-on top of AppConfig->get($opt).
+on top of AppConfig->get($opt). If the option doesn't exist, returns
+C<undef>.
 
 =cut
 
 sub option
 {
     my ($self,$opt) = @_;
```
stdweird: space after `,`

src/main/perl/Application.pm
```diff
@@ -121,20 +121,62 @@ sub username
...
 
-    return $self->{'CONFIG'}->get($opt);
+    if ( $self->option_exists($opt) ) {
+        return $self->{'CONFIG'}->get($opt);
+    }
+
+    return;
```
stdweird: i would support a default value, it's as simple as
```per
    my ($self, $opt, $default) = @_;
    ...
    # last line, even if no default is defined
    return $default;
```

src/main/perl/Application.pm
```diff
@@ -121,20 +121,62 @@ sub username
...
+
+Defines an option and sets its value. If the option was previously
+defined, its value is overwritten. This is a wrapper over C<AppConfig>
+methods to hide the internal implementation of a C<CAF::Application>.
+
+=cut
+
+sub set_option
+{
+    my ($self,$opt,$val) = @_;
```
stdweird: space after `,`

src/test/perl/option.t
```diff
@@ -0,0 +1,38 @@
...
+ok($this_app->option_exists($OPTION1), "Option $OPTION1 defined");
+
+# option() tests
+for my $expected_value ($OPTION_VAL1, $OPTION_VAL2) {
+    $this_app->set_option($OPTION1, $expected_value);
+    my $option_value = $this_app->option($OPTION1);
+    is($option_value, $expected_value, "Option has expected value ($expected_value)");
+}
+my $undef_value = $this_app->option('undefined_option');
+ok(!$undef_value, "undef returned when option is undefined");
```
stdweird: to test undef, you need to use `ok(! defined($undef_value),...`, otherwise this coudl return 0 or empty string or something else

src/test/perl/option.t
```diff
@@ -0,0 +1,38 @@
...
+
+our $this_app = CAF::Application->new('option test');
+
+Readonly my $OPTION1 => 'testopt';
+Readonly my $OPTION_VAL1 => 'value1';
+Readonly my $OPTION_VAL2 => 'value2';
+
+
+# Check that option_exists() returns true if option exists and false otherwise
+ok(!$this_app->option_exists($OPTION1), "Option $OPTION1 doesn't exist");
```
stdweird: test with `ok(!defined()`

src/main/perl/Application.pm
```diff
@@ -121,20 +121,62 @@ sub username
...
+
+    $self->{'CONFIG'}->define($opt);
+    $self->{'CONFIG'}->set($opt, $val);
+
+    return;
```
stdweird: maybe return `$val`? or `SUCCESS`? altough nothing can go wrong, it's odd that we return undef when something is ok

src/main/perl/Application.pm
jouvin: Good suggestion.

src/main/perl/Application.pm
```diff
@@ -121,20 +121,67 @@ sub username
...
+    if ( $self->option_exists($opt) ) {
+        return $self->{'CONFIG'}->get($opt);
+    }
+
+    return $default;
+}
+
+=pod
+
+=item set_option($opt, $val): undef
```
stdweird: `undef` - > `SUCCESS`

src/test/perl/option.t
```diff
@@ -0,0 +1,40 @@
...
+# option() tests for an existing option
+for my $expected_value ($OPTION_VAL1, $OPTION_VAL2) {
+    $this_app->set_option($OPTION1, $expected_value);
+    my $option_value = $this_app->option($OPTION1);
+    is($option_value, $expected_value, "Option has expected value ($expected_value)");
+}
+
+# option() tests for an undefined option
+ok(! defined($this_app->option('undefined_option')), "undef returned when option is undefined (no default value)");
+is($this_app->option('undefined_option', $OPTION_VAL1), $OPTION_VAL1, "default value returned if option is undefined");
```
stdweird: add test that the option still isn't defined after default is used

jouvin: Done

PR #225, opened 2017-03-04
CAF::Path: add methods to manage links

jouvin:
3 methods added to manage symlinks:
- `is_symlink()`: returns true if the argument is a symlink
- `symlink()`: creates a new symlink or updates an existing one. If the symlink path already exists and is not a symlink, returns an error.
- hardlink(): same as `symlink()` for hardlinks.
- has_hardlinks(): test if a file has hardlinks (several entries refering to the same inode)
- is_hardlink(): compare to paths and check if they refer to the same inode
  - Note that there it is impossible to say if a file is a hardlink in itself... A hardlink is just a normal file.

### DIFF
--- a/src/main/perl/Path.pm
+++ b/src/main/perl/Path.pm
@@ -346,6 +346,18 @@ sub any_exists
     return $path && (-e $path || -l $path);
 }
 
+=item is_symlink
+
+Test if C<path> is a symlink.
+
+=cut
+
+sub is_symlink
+{
+    my ($self, $path) = @_;
+    return $path && -l $path;
+}
+
 =item cleanup
 
 cleanup removes C<dest> with backup support.
--- a/src/main/perl/Path.pm
+++ b/src/main/perl/Path.pm
@@ -1,7 +1,7 @@
 #${PMpre} CAF::Path${PMpost}
 
 use CAF::Object qw(SUCCESS CHANGED);
-use LC::Check;
+use LC::Check 1.22;
 use LC::Exception qw (throw_error);
 
 use Readonly;
@@ -357,7 +357,7 @@ sub any_exists
 
 Test if C<path> is a symlink.
 
-Returns true as long as C<path> is a symlink, even though the 
+Returns true as long as C<path> is a symlink, including when the 
 symlink target doesn't exist.
 
 =cut
@@ -573,7 +573,7 @@ returning the standard C<CAF::Path> return values. Every option
 supported by C<LC::Check::link> is supported. C<NoAction>
 flag is handled by C<LC::Check::link> and C<keeps_state> option
 is honored (overrides C<NoAction> if true). One important
-different is the order of the arguments: C<CAF::Path:_make_link>
+difference is the order of the arguments: C<CAF::Path:_make_link>
 and the methods based on it are following the Perl C<symlink>
 (and C<ln> command) argument order.
 
@@ -638,13 +638,14 @@ sub hardlink
 Create a symlink C<link_path> whose target is C<target>.
 
 Returns undef and sets the fail attribute if C<link_path> 
-already exists and is not a symlink, except if this is a file and option C<force> is
-defined and true. If C<link_path> exists and is a symlink, it is updated.
-By default, the target is not required to exist. If you want to
-ensure that it exists, define option C<check> to true.
-Both C<link_path> and C<target> can be relative paths and are
-kept relative in this case. C<link_path> parent directory
-is created if it doesn't exist.
+already exists and is not a symlink, except if this is a file
+and option C<force> is defined and true. If C<link_path> exists
+and is a symlink, it is updated. By default, the target is not 
+required to exist. If you want to ensure that it exists, 
+define option C<check> to true. Both C<link_path> and C<target>
+can be relative paths: C<link_path> is interpreted as relatif
+to the current directory and C<target> is kept relative. 
+C<link_path> parent directory is created if it doesn't exist.
 
 Returns SUCCESS on sucess if the symlink already existed
 with the same target, CHANGED if the symlink was created
@@ -673,6 +674,69 @@ sub symlink
     }
 
     return $self->_make_link($target, $link_path, %opts);
+}
+
+
+=item has_hardlinks
+
+Method that returns the number of hardlinks for C<file>. The number of
+hardlinks is the number of entries referring to the inodes minus 1. If
+C<file> has no hardlink, the return value is 0. If C<file> is not a file,
+the return value is C<undef>.
+
+=cut
+
+sub has_hardlinks
+{
+    my ($self, $file) = @_;
+    $file = $self->_untaint_path($file, "has_hardlinks") || return;
+    
+    if ( ! $self->file_exists($file) && ! $self->is_symlink($file) ) {
+        $self->debug(2, "has_hardlinks(): $file doesn't exist or is not a file");
+        return;
+    }
+    
+    my $nlinks = (lstat($file))[3];
+    $self->debug(2, "Number of links to $file: $nlinks (hardlink if > 2)");
+    return $nlinks ? $nlinks - 1 : 0;
+}
+
+
+=item is_hardlink
+
+This method returns SUCCESS if C<path1> and C<path2> refer to the same file (inode).
+It returns 0 if C<path1> and C<path2> both exist but are different files or are the same path
+and C<undef> if one of the paths doesn't exist or is not a file.
+
+Note: the result returned will be identical whatever is the order of C<path1> and C<path2>
+arguments.
+
+=cut
+
+sub is_hardlink
+{
+    my ($self, $path1, $path2) = @_;
+    $path1 = $self->_untaint_path($path1, "is_hardlink path1") || return;
+    $path2 = $self->_untaint_path($path2, "is_hardlink path2") || return;
+    
+    if ( ! $self->file_exists($path1) && ! $self->is_symlink($path1) ) {
+        $self->debug(2, "is_hardlink(): $path1 doesn't exist or is not a file");
+        return;
+    }
+    if ( ! $self->file_exists($path2) && ! $self->is_symlink($path2) ) {
+        $self->debug(2, "is_hardlink(): $path2 doesn't exist or is not a file");
+        return;
+    }
+    
+    my $link_inode = (lstat($path1))[1];
+    my $target_inode = (lstat($path2))[1];
+
+    $self->debug(2, "Comparing $path1 inode ($link_inode) and $path2 inode ($target_inode)");
+    if ( ($link_inode == $target_inode) && ($path1 ne $path2)  ) {
+        return SUCCESS;
+    } else {
+        return 0;
+    }
 }
 
 
--- a/src/main/perl/Path.pm
+++ b/src/main/perl/Path.pm
@@ -28,6 +28,7 @@ Readonly::Hash my %CLEANUP_DISPATCH => {
 # Use dispatch table instead of non-strict function by variable call
 Readonly::Hash my %LC_CHECK_DISPATCH => {
     directory => \&LC::Check::directory,
+    link => \&LC::Check::link,
     status => \&LC::Check::status,
 };
 
@@ -356,6 +357,9 @@ sub any_exists
 
 Test if C<path> is a symlink.
 
+Returns true as long as C<path> is a symlink, even though the 
+symlink target doesn't exist.
+
 =cut
 
 sub is_symlink
@@ -559,6 +563,116 @@ sub directory
     # A new directory always implies something changed
     my $changed = ($newdir || $status == CHANGED) ? CHANGED : SUCCESS;
     return dualvar( $changed, $directory);
+}
+
+
+=item _make_link
+
+This method is mainly a wrapper over C<LC::Check::link>
+returning the standard C<CAF::Path> return values. Every option
+supported by C<LC::Check::link> is supported. C<NoAction>
+flag is handled by C<LC::Check::link> and C<keeps_state> option
+is honored (overrides C<NoAction> if true). One important
+different is the order of the arguments: C<CAF::Path:_make_link>
+and the methods based on it are following the Perl C<symlink>
+(and C<ln> command) argument order.
+
+This is an internal method, not supposed to be called directly.
+Either call C<symlink> or C<hardlink> public methods instead.
+
+=cut
+
+sub _make_link
+{
+    my ($self, $target, $link_path, %opts) = @_;
+    my $link_type = $opts{hard} ? "hardlink" : "symlink";
+
+    $link_path = $self->_untaint_path($link_path, $link_type) || return;
+    $target = $self->_untaint_path($target, $link_type) || return;
+
+    $self->_reset_exception_fail($link_type);
+
+    my $status = $self->LC_Check('link', [$link_path, $target], \%opts);
+            
+    if ( defined($status) ) {
+        return ($status ? CHANGED : SUCCESS);
+    } else {
+        return;
+    }
+}
+
+=item hardlink
+
+Create a hardlink C<link_path> whose target is C<target>.
+
+On failure, returns undef and sets the fail attribute.
+If C<link_path> exists and is a file, it is updated. 
+C<target> must exist (C<check> flag available in symlink()
+is ignored for hardlinks) and it must reside in the same 
+filesystem as C<link_path>. If C<target_path> is a
+relative path, it is interpreted from the current directory.
+C<link_name> parent directory is created if it doesn't exist.
+
+Returns SUCCESS on sucess if the hardlink already existed
+with the same target, CHANGED if the hardlink was created
+or updated, undef otherwise.
+
+This method relies on C<_make_link> method to do the real work,
+after enforcing the option saying that it is a hardlink.
+
+=cut
+
+sub hardlink
+{
+    my ($self, $target, $link_path, %opts) = @_;
+
+    # Option passed to LC::Check::link to indicate a hardlink
+    $opts{hard} = 1;
+
+    return $self->_make_link($target, $link_path, %opts);
+}
+
+
+=item symlink
+
+Create a symlink C<link_path> whose target is C<target>.
+
+Returns undef and sets the fail attribute if C<link_path> 
+already exists and is not a symlink, except if this is a file and option C<force> is
+defined and true. If C<link_path> exists and is a symlink, it is updated.
+By default, the target is not required to exist. If you want to
+ensure that it exists, define option C<check> to true.
+Both C<link_path> and C<target> can be relative paths and are
+kept relative in this case. C<link_path> parent directory
+is created if it doesn't exist.
+
+Returns SUCCESS on sucess if the symlink already existed
+with the same target, CHANGED if the symlink was created
+or updated, undef otherwise.
+
+This method relies on C<_make_link> method to do the real work,
+after enforcing the option saying that it is a symlink.
+
+=cut
+
+sub symlink
+{
+    my ($self, $target, $link_path, %opts) = @_;
+
+    # Option passed to LC::Check::link to indicate a symlink
+    $opts{hard} = 0;
+    
+    # LC::Check::symlink() expects an option 'nocheck' but CAF::Path::symlink exposes
+    # an option 'check', as the default in CAF::Path::symlink is not to check the
+    # target existence. Convert it to 'nocheck'.
+    if ( defined($opts{check}) ) {
+        $opts{nocheck} = ! $opts{check};
+        delete $opts{check};
+    } else {
+        $opts{nocheck} = 1;
+    }
+
+    return $self->_make_link($target, $link_path, %opts);
 }
 
 
--- a/src/main/perl/Path.pm
+++ b/src/main/perl/Path.pm
@@ -196,8 +196,10 @@ sub _function_catch
 
 Run function reference C<funcref> with arrayref C<argsref> and hashref C<optsref>.
 
-Return and set fail attribute with C<failmsg> on die, verbose C<msg> on success
-(resp. $@ and stringified result are appended).
+Return and set fail attribute with C<failmsg> on die or an error (C<undef> returned
+by C<funcref>), or print (at verbose level) C<msg> on success (respectively $@ and 
+stringified result are appended). Note that C<_safe_eval> doesn't work with functions
+that don't return a defined value when they succeed.
 
 Resets previous exceptions and/or fail attribute
 
@@ -209,22 +211,26 @@ sub _safe_eval
 
     $self->_reset_exception_fail('_safe_eval');
 
-    my ($res, @args, %opts);
+    my (@args, %opts);
     @args = @$argsref if $argsref;
     %opts = %$optsref if $optsref;
 
     local $@;
-    eval {
-        $res = $funcref->(@args, %opts);
-    };
+    my $res = eval {
+                    $funcref->(@args, %opts);
+                   };
 
-    if ($@) {
-        chomp($@);
-        return $self->fail("$failmsg: $@");
-    } else {
-        my $res_txt = defined($res) ? "$res" : '<undef>';
-        chomp($res);
+    # $res is undef if there is a syntax or runtime error or if the evaluated
+    # function returns undef (interpreted as a function error).
+    if ( defined($res) ) {
         $self->verbose("$msg: $res");
+    } else {
+        my $err_msg = '';
+        if ($@) {
+            chomp($@);
+            $err_msg = ": $@";
+        }
+        return $self->fail("$failmsg$err_msg");
     }
 
     return $res;
--- a/src/test/perl/path.t
+++ b/src/test/perl/path.t
@@ -102,19 +102,21 @@ sub verify_exception
 {
     my ($msg, $fail, $expected_reset, $noreset) = @_;
     $expected_reset = 1 if (! defined($expected_reset));
-    is($exception_reset, $expected_reset, "exception_reset called $expected_reset after $msg");
+    is($exception_reset, $expected_reset, "_reset_exception_fail called $expected_reset times after $msg");
     if ($noreset) {
         ok($ec_check->error(), "Error not reset after $msg");
     } else {
         ok(! $ec_check->error(), "Error reset after $msg");
     };
-    if ($noreset) {
+    if ($noreset && defined($mc->{fail})) {
         like($mc->{fail}, qr{^origfailure }, "Fail attribute matches originalfailure on noreset after $msg");
-    } elsif ($fail) {
+    } elsif ($fail && defined($mc->{fail})) {
         like($mc->{fail}, qr{$fail}, "Fail attribute matches $fail after $msg");
         unlike($mc->{fail}, qr{origfailure}, "original fail attribute reset");
-    } else {
+    } elsif ( ! $noreset ) {
         ok(! defined($mc->{fail}), "Fail attribute reset after $msg");
+    } else {
+        ok(0, "internal test error: unexpected undefined fail attribute") if (! defined($mc->{fail}));
     };
 };
 
--- a/src/test/perl/path.t
+++ b/src/test/perl/path.t
@@ -290,20 +290,24 @@ rmtree if -d $basetest;
 ok(! $mc->directory_exists($basetest), "directory_exists false on missing directory");
 ok(! $mc->file_exists($basetest), "file_exists false on missing directory");
 ok(! $mc->any_exists($basetest), "any_exists false on missing directory");
+ok(! $mc->is_symlink($basetest), "is_symlink false on missing directory");
 
 ok(! $mc->directory_exists($basetestfile), "directory_exists false on missing file");
 ok(! $mc->file_exists($basetestfile), "file_exists false on missing file");
 ok(! $mc->any_exists($basetestfile), "any_exists false on missing file");
+ok(! $mc->is_symlink($basetestfile), "is_symlink false on missing file");
 
 makefile($basetestfile);
 
 ok($mc->directory_exists($basetest), "directory_exists true on created directory");
 ok($mc->any_exists($basetest), "any_exists true on created directory");
 ok(! $mc->file_exists($basetest), "file_exists false on created directory");
+ok(! $mc->is_symlink($basetest), "is_symlink false on created directory");
 
 ok(! $mc->directory_exists($basetestfile), "directory_exists false on created file");
 ok($mc->any_exists($basetestfile), "any_exists true on created file");
 ok($mc->file_exists($basetestfile), "file_exists true on created file");
+ok(! $mc->is_symlink($basetestfile), "is_symlink false on created file");
 
 # Test (broken) symlink and _exsists methods
 
@@ -315,14 +319,17 @@ ok(symlink("tgtdir/tgtfile", $filelink), "file symlink created");
 ok(! $mc->directory_exists($brokenlink), "directory_exists false on brokenlink");
 ok(! $mc->file_exists($brokenlink), "file_exists false on brokenlink");
 ok($mc->any_exists($brokenlink), "any_exists true on brokenlink");
+ok($mc->is_symlink($brokenlink), "is_symlink true on brokenlink");
 
 ok($mc->directory_exists($dirlink), "directory_exists true on dirlink");
 ok(! $mc->file_exists($dirlink), "file_exists false on dirlink");
 ok($mc->any_exists($dirlink), "any_exists true on dirlink");
+ok($mc->is_symlink($dirlink), "is_symlink true on dirlink");
 
 ok(! $mc->directory_exists($filelink), "directory_exists false on filelink");
 ok($mc->file_exists($filelink), "file_exists true on filelink");
 ok($mc->any_exists($filelink), "any_exists true on filelink");
+ok($mc->is_symlink($filelink), "is_symlink true on filelink");
 
 
 # noreset=1
--- a/src/test/perl/path.t
+++ b/src/test/perl/path.t
@@ -11,6 +11,7 @@ use LC::Exception qw (throw_error);
 
 use Test::More;
 use Test::MockModule;
+use Cwd;
 
 use CAF::Object qw(SUCCESS CHANGED);
 use CAF::Path;
@@ -81,6 +82,11 @@ ok(! $mc->_get_noaction(1), "_get_noaction returns false with CAF::Object::NoAct
 
 my $exception_reset = 0;
 
+# init_exception() and verify_exception() functions work in pair. They allow to register a message
+# in 'fail' attribute at the beginning of a test section and to verify if new (unexpected) exceptions
+# where raised during the test section. To reset the 'fail' attribute after verify_exception(),
+# call _reset_exception_fail(). init_exception() implicitely resets the 'fail' attribute and also
+# reset to 0 the count of calls to _reset_exception_fail().
 sub init_exception
 {
     my ($msg) = @_;
@@ -135,6 +141,21 @@ $mock->mock('_reset_exception_fail', sub {
     return &$init(@_);
 });
 
+# Mocked symlink() and hardlink() to count calls
+my $symlink_call_count = 0;
+my $hardlink_call_count = 0;
+
+$mock->mock('symlink', sub {
+    $symlink_call_count += 1;
+    my $symlink_orig = $mock->original('symlink');
+    return &$symlink_orig(@_);
+});
+
+$mock->mock('hardlink', sub {
+    $hardlink_call_count += 1;
+    my $hardlink_orig = $mock->original('hardlink');
+    return &$hardlink_orig(@_);
+});
 
 =head2 _function_catch
 
@@ -279,11 +300,12 @@ $mc->{fail} = undef;
 is($mc->_untaint_path("abc", "ok"), "abc", "untaint ok");
 ok(! defined($mc->{fail}), "no fail attribute set with ok path");
 
-=head2 directory/file/any exists
+
+=head2 directory/file/any exists on files and directories
 
 =cut
 
-init_exception("existence tests");
+init_exception("existence tests (file/directory)");
 
 # Tests without NoAction
 $CAF::Object::NoAction = 0;
@@ -311,12 +333,113 @@ ok($mc->any_exists($basetestfile), "any_exists true on created file");
 ok($mc->file_exists($basetestfile), "file_exists true on created file");
 ok(! $mc->is_symlink($basetestfile), "is_symlink false on created file");
 
-# Test (broken) symlink and _exsists methods
+# noreset=1
+verify_exception("existence tests (file/directory)", undef, 0, 1);
+ok($mc->_reset_exception_fail(), "_reset_exception_fail after existence tests (file/directory)");
 
-ok(symlink("really_really_missing", $brokenlink), "broken symlink created");
-makefile("$basetest/tgtdir/tgtfile");
-ok(symlink("tgtdir", $dirlink), "directory symlink created");
-ok(symlink("tgtdir/tgtfile", $filelink), "file symlink created");
+
+=head2 symlink/hardlink creation/update/test
+
+=cut
+
+# Test symlink creation
+
+# Function to do the ok()/is() pair, taking into account NoAction flag
+sub check_symlink {
+    my ($mc, $target, $link_path, $expected_status) = @_;
+    my $noaction = $mc->_get_noaction();
+
+    my $ok_msg;
+    if ( $noaction ) {
+        $ok_msg = "$link_path symlink not created (NoAction set)";
+    } else {
+        $ok_msg = "$link_path is " . ($expected_status == CHANGED ? "" : "already ") . "a symlink";
+    }
+    my $target_msg = "$link_path symlink has the expected " . ($expected_status == CHANGED ? "changed " : "") . "target ($target)";
+
+    # Test if symlink was created or not according to NoAction flag
+    my $ok_condition;
+    if ( $noaction ) {
+        $ok_condition = ! $mc->any_exists($link_path);
+    } else {
+        $ok_condition = $mc->is_symlink($link_path);
+    }
+    ok($ok_condition, $ok_msg);
+
+    is(readlink($link_path), $target, $target_msg) unless $noaction;
+};
+
+init_exception("symlink tests");
+
+rmtree ($basetest) if -d $basetest;
+my $target_directory = "tgtdir";
+my $target_file1 = "tgtfile1";
+my $target_file2 = "$basetest/$target_directory/tgtfile2";
+makefile("$basetest/$target_file1");
+makefile($target_file2);
+my %opts;
+
+# Symlink creations
+for $CAF::Object::NoAction (1,0) {
+    is($mc->symlink($target_directory, $dirlink), CHANGED, "directory symlink created");
+    check_symlink($mc, $target_directory, $dirlink, CHANGED);
+    is($mc->symlink($target_file2, $filelink), CHANGED, "file symlink created");
+    check_symlink($mc, $target_file2, $filelink, CHANGED);
+}
+
+# Valid symlink updates
+# The following tests make no sens if NoAction is true (require the previous creations to occur)
+is($mc->symlink($target_file2, $filelink), SUCCESS, "file symlink already exists: nothing done");
+check_symlink($mc, $target_file2, $filelink, SUCCESS);
+is($mc->symlink($target_file1, $filelink), CHANGED, "file symlink updated");
+check_symlink($mc, $target_file1, $filelink, CHANGED);
+ok(! readlink($target_file1), "$target_file1 exists and is a file");
+my $link_status = $mc->symlink($target_file1, $target_file2);
+ok(! defined($link_status), "symlink failed: existing file not replaced by a symlink");
+is($mc->{fail},
+   "*** cannot symlink $target_file2: it is not an existing symlink",
+   "fail attribute set after symlink failure (existing file not replaced by a symlink)");
+ok($mc->file_exists($target_file2) && ! $mc->is_symlink($target_file2), "File $target_file2 has not be replaced by a symlink");
+$opts{force} = 1;
+$CAF::Object::NoAction = 1;
+is($mc->symlink($target_file1, $target_file2, %opts), CHANGED, "existing file replaced by a symlink (force option set)");
+ok($mc->file_exists($target_file2), "File $target_file2 not replaced by a symlink with 'force' option (NoAction set)");
+$CAF::Object::NoAction = 0;
+is($mc->symlink($target_file1, $target_file2, %opts), CHANGED, "existing file replaced by a symlink (force option set)");
+check_symlink($mc, $target_file1, $target_file2, CHANGED);
+
+# Invalid symlink updates
+my $test_directory = "$basetest/$target_directory";
+$link_status = $mc->symlink($target_file1, "$test_directory", %opts);
+ok (! defined($link_status), "directory not replaced by a symlink (force option set)");
+is($mc->{fail},
+   "*** cannot symlink $test_directory: it is not an existing symlink",
+   "fail attribute set after symlink failure (existing directory not replaced by a symlink)");
+ok($mc->directory_exists($test_directory) && ! $mc->is_symlink($test_directory),
+   "Directory $test_directory has not be replaced by a symlink");
+
+# Broken symlinks with and without 'check' option
+$opts{check} = 1;
+ok(! $mc->symlink("really_really_missing", $brokenlink, %opts), "broken symlink not created (target existence enforced)");
+ok(! -e $brokenlink && ! -l $brokenlink, "Broken link has not been created");
+$opts{check} = 0;
+is($mc->symlink("really_missing", $brokenlink), CHANGED, "broken symlink created");
+ok($mc->is_symlink($brokenlink), "Broken link has been created (target check disabled by check=0)");
+is(readlink($brokenlink), "really_missing", "Broken link has the expected target by check=0");
+delete $opts{check};
+is($mc->symlink("really_really_missing", $brokenlink), CHANGED, "broken symlink updated");
+ok($mc->is_symlink($brokenlink), "Broken link has been updated (target check disabled by 'check' undefined)");
+is(readlink($brokenlink), "really_really_missing", "Broken link has the expected target by 'check' undefined");
+
+# noreset=0
+verify_exception("symlink tests", "Failed to create symlink", $symlink_call_count * 2, 0);
+ok($mc->_reset_exception_fail(), "_reset_exception_fail after symlink tests");
+
+
+# Test xxx_exists methods with symlinks
+# Needs symlinks created in previous step (symlink creation/update)
+
+init_exception("existence tests (symlinks)");
 
 ok(! $mc->directory_exists($brokenlink), "directory_exists false on brokenlink");
 ok(! $mc->file_exists($brokenlink), "file_exists false on brokenlink");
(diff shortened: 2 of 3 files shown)


src/main/perl/Path.pm
```diff
@@ -346,6 +346,54 @@ sub any_exists
     return $path && (-e $path || -l $path);
 }
 
+=item is_symlink
+
+Test if C<path> is a symlink.
```
stdweird: pod should mention this reutrn true for broken symlink

src/main/perl/Path.pm
```diff
@@ -346,6 +346,54 @@ sub any_exists
...
+Create a symlink C<path> whose target is C<target>.
+
+Returns an error if C<path> already exists and is not a 
+symlink. If it exists and is a symlink, it is updated.
+
+Returns SUCCESS on sucess, undef otherwise.
+
+=cut
+
+sub make_symlink
```
stdweird: like all other methods in CAF::Path, this has to support NoAction (incl unittests). i'd suggest to check the LC code and use that for now. we can replace LC later.

src/main/perl/Path.pm
```diff
@@ -346,6 +346,54 @@ sub any_exists
...
+{
+    my ($self, $target, $path) = @_;
+    my $status = SUCCESS;     # Assume success
+    unless ($path) {
```
stdweird: i propose to keep the same style as the rest of CAF::Path, so don't check this. lets assume that people who use this know what the are doing.

src/main/perl/Path.pm
```diff
@@ -346,6 +346,54 @@ sub any_exists
...
+        $self->error("create_symlink(): 'path' argument missing");
+    }
+    unless ($target) {
+        $self->error("create_symlink(): 'target' argument missing");
+    }
+
```
stdweird: you have too untaint path and target, look at eg the directory code (you can ignore the symlik part)

src/main/perl/Path.pm
```diff
@@ -346,6 +346,54 @@ sub any_exists
...
+    if ( $self->is_symlink($path) ) {
+        $self->debug(1, "Removing symlink $path");
+        unlink ($path);
```
stdweird: use CAF::Path methods to remove things.

src/main/perl/Path.pm
```diff
@@ -346,6 +346,54 @@ sub any_exists
...
+    my $status = SUCCESS;     # Assume success
+    unless ($path) {
+        $self->error("create_symlink(): 'path' argument missing");
+    }
+    unless ($target) {
+        $self->error("create_symlink(): 'target' argument missing");
+    }
+
+    if ( $self->is_symlink($path) ) {
+        $self->debug(1, "Removing symlink $path");
```
stdweird: should only remove of something target is different

src/main/perl/Path.pm
```diff
@@ -346,6 +346,54 @@ sub any_exists
...
+        unlink ($path);
+    } elsif ( $self->any_exists($path) ) {
+        $self->error("$path already exists and is not a symlink");
+        return;
+    }
+
+    $status = symlink $target, $path;
```
stdweird: use LC method

src/main/perl/Path.pm
```diff
@@ -346,6 +346,54 @@ sub any_exists
...
+    $status = SUCCESS if $status;
```
stdweird: success should be eturned when nothing has changed (eg link already existed and pointed to same target).
return CHANGED if a link was actually created

src/main/perl/Path.pm
stdweird: i would also rename it to just `symlink`. the result of this method should be a symlink, it's possible none should be created.

src/main/perl/Path.pm
```diff
@@ -346,6 +346,54 @@ sub any_exists
...
+    }
+    unless ($target) {
+        $self->error("create_symlink(): 'target' argument missing");
+    }
+
+    if ( $self->is_symlink($path) ) {
+        $self->debug(1, "Removing symlink $path");
+        unlink ($path);
+    } elsif ( $self->any_exists($path) ) {
+        $self->error("$path already exists and is not a symlink");
```
stdweird: report failures with `fail`, eg. `return $self->fail("$path already exists and is not a symlink");`
in particular, CAF should try to minimise the reported errors. there might be a good reason that something fails, it's up to the component to promote a failure as an error (since any error means component failed).

stdweird:
@jouvin please have a look at eg the directory code. try to use the wrapped LC calls or existing CAF::Path methods.
there has to be support for NoAction and there have to be unittests for NoAction

src/main/perl/Path.pm
stdweird: and make (or try to make) the parent dir first

jouvin:
@stdweird thanks for all the remarks, I'll try to improve the code later today...

src/main/perl/Path.pm
jouvin: I looked at `directory()` and I don't see `NoAction` being handled (`NoAction` is handled only for temporary directories)... This may explain https://github.com/quattor/maven-tools/issues/139...

stdweird: NoAction is handled in the LC call, the Noaction flag is passed in https://github.com/quattor/CAF/blob/master/src/main/perl/Path.pm#L252; only functionality not in LC requires explicit NoAction code.
best point to start is to look how symlinking is handled in LC.

jouvin: This is what I was doing and I have the feeling that the method in `CAF::Path` can be almost as simple as calling `LC::Check::symlink`... Good news!

jouvin:
@stdweird @jrha @ned21 any opinion on the argument order (link path versus target)? Perl `symlink` is following `ln` command with target first, link second. `LC::Check::symlink` is following the order I had in mind initially: link first, target second. As `CAF::Path::symlink` is mainly a wrapper over  `LC::Check::symlink`, I'm inclined to follow its order but I don't have a strong opinion. And I am happy to change if you think the Perl `symlink` order is less confusion/more appropriate...

jouvin:
Reimplemented over `LC::Check:link()` as suggested by @stdweird. Some problems in tests, to be fixed later...

ned21:
@jouvin - tough question! Following LC::Check::symlink makes it easier to refactor code in the short term but long term following ```symlink()``` is better for everyone and less confusing. Since I think we want to take a long term view, I vote for following the Perl and UNIX symlink ordering.

jouvin:
I think the code is now in pretty good shape. We may just need to adjust the argument order and in fact part of the refactoring we want to do after having these methods is replacing `symlink` by `CAF::Path`. For this @ned21 proposal is probably less confusing, I agree...

I am still struggling with the tests: @stdweird `verify_exception()` is failing and I don't really understand what it is supposed to do (and I have not seen any comment in the code to help with it!). Any explanation much appreciated! Once tests are working for `symlink()`, I still need to add tests for `hardlink()` and tests for `NoAction` but it should be easy.

src/main/perl/Path.pm
```diff
@@ -544,6 +559,99 @@ sub directory
...
+
+This in internal method, not supposed to be called directly.
+Either call C<symlink> or C<hardlink> public methods instead.
+
+=cut
+
+sub _make_link
+{
+    my ($self, $path, $target, %opts) = @_;
+    $path = $self->_untaint_path($path, "symlink") || return;
```
stdweird: should be `symlink path` as message, and why isn't the target untainted?

src/main/perl/Path.pm
```diff
@@ -544,6 +559,99 @@ sub directory
...
+
+    $self->_reset_exception_fail('symlink');
+
+    my $status = LC::Check::link($path, $target, %opts);
```
stdweird: use the wrapper to call `LC`, eg https://github.com/quattor/CAF/blob/master/src/main/perl/Path.pm#L532

src/main/perl/Path.pm
```diff
@@ -544,6 +559,99 @@ sub directory
...
+
+    if ( defined($status) ) {
+        return ($status ? CHANGED : SUCCESS);
+    } else {
+        my $link_type = $opts{hard} ? "hardlink" : "symlink";
+        $self->fail("Failed to create $link_type $path (target=$target)");
```
stdweird: `return $self->fail()`

src/main/perl/Path.pm
```diff
@@ -544,6 +559,99 @@ sub directory
...
+        return;
+    }
+}
+
+=item hardlink
+
+Create a hardlink C<path> whose target is C<target>.
+
+Returns an error if C<path> already exists and is not a
```
stdweird: `Returns an error` -> `Returns undef on failure and sets the fail attribute`

src/main/perl/Path.pm
```diff
@@ -544,6 +559,99 @@ sub directory
...
+
+    return $status;
+}
+
+
+=item symlink
+
+Create a symlink C<path> whose target is C<target>.
+
+Returns an error if C<path> already exists and is not a
```
stdweird: same `Returns error` as above

src/main/perl/Path.pm
```diff
@@ -544,6 +559,99 @@ sub directory
...
+
+
+=item symlink
+
+Create a symlink C<path> whose target is C<target>.
+
+Returns an error if C<path> already exists and is not a 
+symlink, except if this is a file and option C<force> is
+defined and true. If C<path> exists and is a symlink, it is updated.
+By default, the target is not required to exist. If you want to
```
stdweird: this is odd. `the target is not required to exist` suggest something is not checked; but to enable the check you have to set `nocheck` to true?

i would suspect to have to set a variable `check` to true to enable an existence check, it makes much more sense.

src/main/perl/Path.pm
```diff
@@ -544,6 +559,99 @@ sub directory
...
+
+=cut
+
+sub symlink
+{
+    my ($self, $path, $target, %opts) = @_;
+
+    $opts{hard} = 0;
+    # Disabled the target existence check if 'nocheck' option not specified
+    unless ( defined($opts{nocheck}) ) {
```
stdweird: try to avoid `unless`, it's very confusing to read, esp in combination with variable names that suggest an inversed logic.
you also do not check the value of `nocheck`?

src/main/perl/Path.pm
```diff
@@ -544,6 +559,99 @@ sub directory
...
+        $opts{nocheck} = 1;
+    }
+    $self->verbose("nocheck=$opts{nocheck}");
```
stdweird: this message needs some more explanation, like symlink, path/target

src/main/perl/Path.pm
```diff
@@ -544,6 +559,99 @@ sub directory
...
+    my $status = $self->_make_link($path, $target, %opts);
```
stdweird: simply `return $self->_make_links`

src/test/perl/path.t
```diff
@@ -108,9 +108,9 @@ sub verify_exception
     } else {
         ok(! $ec_check->error(), "Error reset after $msg");
     };
-    if ($noreset) {
+    if ($noreset && $mc->{fail}) {
```
stdweird: please don't modify existing tests. using the `LC` wrapper is probably enough to make this pass.

src/test/perl/path.t
```diff
@@ -290,39 +290,68 @@ rmtree if -d $basetest;
...
+
+# Test symlink creation
+
+my $target_directory = "tgtdir";
+my $target_file1 = "tgtfile1";
+my $target_file2 = "$basetest/$target_directory/tgtfile2";
+makefile("$basetest/$target_file1");
+makefile($target_file2);
+my %opts;
+is($mc->symlink($dirlink, $target_directory), CHANGED, "directory symlink created");
```
stdweird: add a symlink test ` $mc->is_symlink(` for every symlink created

stdweird: the main reason for this is so you can use the same/similar code block under noaction with minimal changes

src/test/perl/path.t
```diff
@@ -290,39 +290,68 @@ rmtree if -d $basetest;
...
+is($mc->symlink($filelink, $target_file2), CHANGED, "file symlink created");
+is($mc->symlink($filelink, $target_file2), SUCCESS, "file symlink already exists: nothing done");
+is($mc->symlink($filelink, $target_file1), CHANGED, "file symlink already exists: nothing done");
+my $link_status = $mc->symlink($target_file2, $target_file1);
+ok(! $link_status, "existing file not replaced by a symlink");
```
stdweird: it has to return `undef`, so use `ok(!defined($link_status))`

stdweird: and the message should also have `symlink failed failed, reason `

src/test/perl/path.t
```diff
@@ -290,39 +290,68 @@ rmtree if -d $basetest;
...
+is($mc->{fail}, "Failed to create symlink $target_file2 (target=$target_file1)");
```
stdweird: add a message to the test `symlink failed because existing file not replaced by a symlink sets correct fail attrribute`

src/test/perl/path.t
```diff
@@ -290,39 +290,68 @@ rmtree if -d $basetest;
...
+$opts{force} = 1;
+is($mc->symlink($target_file2, $target_file1, %opts), CHANGED, "existing file replaced by a symlink (force option set)");
+ok(! $mc->symlink("$basetest/$target_directory", $target_file1, %opts), "directory not replaced by a symlink (force option set)");
```
stdweird: ok, i reread the pod, and this is implied, but it would be better to make it explicit in the pod (or use the not replacing of a directory as an exmaple in the pod).

src/test/perl/path.t
```diff
@@ -290,39 +290,68 @@ rmtree if -d $basetest;
...
+$opts{nocheck} = 0;
```
stdweird: ahhh, this is confusing. so the pod is wrong?

src/test/perl/path.t
stdweird: same remark as above, here (and everywhere else) add an explicit test that nothing was created when it shouldn't have.

src/test/perl/path.t
```diff
@@ -290,39 +290,68 @@ rmtree if -d $basetest;
...
+ok(! $mc->symlink($brokenlink, "really_really_missing", %opts), "broken symlink not created (target existence enforced)");
+delete $opts{nocheck};
+is($mc->symlink($brokenlink, "really_missing"), CHANGED, "broken symlink created");
```
stdweird: you have to test this, either use `!-e $brokenlink && -l $brokenlink` or a combination of CAF::Path methods

src/main/perl/Path.pm
jouvin: I agree. `nocheck` is what is used in `LC::Check::link`, this is not my choice! We can probably implement a `check` option in `CAF::Path` and convert it to `nocheck` before calling `LC::Check::link`.

src/main/perl/Path.pm
jouvin: Mistake of mine! I thought I did it!

src/test/perl/path.t
jouvin: I didn't really modify existing tests but made a commit explictly dedicated to this as I think the test is wrong... I don't like that you may want to test/print something undefined. It doesn't make sense.

src/test/perl/path.t
jouvin: Not sure to understand... Why `is_symlink()` is improving the ability to use the same code with `NoAction`?

jouvin:
@stdweird see https://github.com/quattor/CAF/pull/225#issuecomment-284270389, I'm still interested by a short explanation about `verify_exceptions()` and what it is supposed to do...

src/main/perl/Path.pm
jouvin: Sorry a left-over from my tests...

src/test/perl/path.t
stdweird: it changes the logic of the test, from one condition to two. if `$mc->{fail}` is not defined, something is wrong and test will fail, as it is supposed to.

jouvin: I understand that my problem was coming from the fact I was not using `_safe_eval()`. Nevertheless, I think we should protect in some way against an `undef` argument rather than adding a warning to the error...

BTW, I think there is a typo at the end of `_safe_eval()`. Should be `chomp $res_text` and `verbose()` should print `$res_text`, shouldn't it?

src/test/perl/path.t
stdweird: the code block for NoAction will be the the smae, except where you now have/would have `ok($mc->is_symlink,..)`, you wil have `ok(!$mc->is_symlink`; since noaction is not supposed to create actual symlink (unless keepsstate is set).

src/main/perl/Path.pm
stdweird: please use `check`, it's not because we use LC internally (for now), we will do so forever or that we should use the same option names or argument order. now is the time to try to make a clean API

src/test/perl/path.t
stdweird: nice catch on the `res_txt`! (i'm waiting for https://github.com/quattor/maven-tools/issues/130 to catch these sort of errors)

if you want to, you can make the tests more robust, but not by changing the logic like this.

jouvin: I don't think it is changing the logic, at least it was not the goal! I need to reread what I did may be too quickly...

jouvin: You are right that `$noreset` was not honored when '$mc->{fail}` was undefined. Should be fixed now and I really think that nothing is changed in the logic, just meaningless `like()` are not executed. @stdweird do you agree?

src/main/perl/Path.pm
jouvin: Completely agreed! Done!

jouvin:
Should be really better now (as usual!) 😄 At this point, the PR is still failing tests because of the problem described in https://github.com/quattor/CAF/pull/225#issuecomment-284270389 but the following problems have been addressed:
- Argument order is the same as symlink
- `_safe_eval()` has been fixed: for me, the way to detect failures were only detecting run time errors, not functions returning a failed status (`undef`). See [eval](http://perldoc.perl.org/functions/eval.html) docucmentation. As a result, `$self->{fail}` was not set. New code is working as expected with all the previously existing tests. @stdweird please validate! (this is in a separate commit).

Still to be done in unit tests:
- unit tests for `NoAction`
- Check the result of actions is what it is said to be by checking that the object is really created or really not created/updated.

I'd like to get feedback on what is available first...

src/main/perl/Path.pm
```diff
@@ -214,17 +216,20 @@ sub _safe_eval
     %opts = %$optsref if $optsref;
 
     local $@;
-    eval {
-        $res = $funcref->(@args, %opts);
-    };
+    my $status = eval {
```
stdweird: i had to read up what eval returned, but it is the same thing as `$res`.

src/main/perl/Path.pm
```diff
@@ -214,17 +216,20 @@ sub _safe_eval
...
+                      };
 
-    if ($@) {
-        chomp($@);
-        return $self->fail("$failmsg: $@");
-    } else {
-        my $res_txt = defined($res) ? "$res" : '<undef>';
-        chomp($res);
+    # $status is undef if there is a syntax or runtime error or if $res is undef.
+    if ( defined($status) ) {
```
stdweird: why is this better than checking `$@`?

src/main/perl/Path.pm
```diff
@@ -544,6 +564,107 @@ sub directory
...
+{
+    my ($self, $target, $link_path, %opts) = @_;
+    my $link_type = $opts{hard} ? "hardlink" : "symlink";
+
+    $link_path = $self->_untaint_path($link_path, $link_type) || return;
+    $target = $self->_untaint_path($target, $link_type) || return;
+
+    $self->_reset_exception_fail($link_type);
+
+    my $status = $self->_safe_eval(\&LC::Check::link,
```
stdweird: why do you use the `LC_Check` wrapper?

src/test/perl/path.t
stdweird: not really. now there's no catch-all else block anymore. in particular i don't see what you gain with this. what did the error message look like when `$mc->{fail}` is undefined?

src/test/perl/path.t
```diff
@@ -290,45 +292,103 @@ rmtree if -d $basetest;
...
+$opts{check} = 1;
+ok(! $mc->symlink("really_really_missing", $brokenlink, %opts), "broken symlink not created (target existence enforced)");
+delete $opts{nocheck};
+is($mc->symlink("really_missing", $brokenlink), CHANGED, "broken symlink created");
+
+# Test hardlink creation
+my $cwd = cwd();
+my $hardlink1 = "$basetest/$target_directory/hardlink1";
+my $hardlink2 = "$basetest/$target_directory/hardlink2";
+# hardlink target must be an absolute path or the link path directory is prepended
```
stdweird: this "feature" should be documented in the pod. it's certainly new to me. and you should add a test for this behaviour; these details.

src/test/perl/path.t
```diff
@@ -290,45 +292,103 @@ rmtree if -d $basetest;
...
+
+my $target_directory = "tgtdir";
+my $target_file1 = "tgtfile1";
+my $target_file2 = "$basetest/$target_directory/tgtfile2";
+makefile("$basetest/$target_file1");
+makefile($target_file2);
+my %opts;
+is($mc->symlink($target_directory, $dirlink), CHANGED, "directory symlink created");
+is($mc->symlink($target_file2, $filelink), CHANGED, "file symlink created");
+is($mc->symlink($target_file2, $filelink), SUCCESS, "file symlink already exists: nothing done");
```
stdweird: add tests that verify the claimed behaviour. there really cannot be any doubt about what goes on here. it is in particular important that these tests can be fully trusted if we ever replace `LC::Check` with something else.

src/test/perl/path.t
```diff
@@ -290,45 +292,103 @@ rmtree if -d $basetest;
...
+# Test hardlink creation
+my $cwd = cwd();
+my $hardlink1 = "$basetest/$target_directory/hardlink1";
+my $hardlink2 = "$basetest/$target_directory/hardlink2";
+# hardlink target must be an absolute path or the link path directory is prepended
+$target_file1 = "$cwd/$basetest/tgtfile1";
+my $target_file3 = "$cwd/$basetest/$target_directory/tgtfile3";
+makefile($target_file1);
+makefile($target_file3);
+is($mc->hardlink($target_file1, $hardlink1), CHANGED, "hardlink created");
```
stdweird: same remark as before: test the claims made by the code by verifying that there really is a file `$hardlink` and that is really is a hardlink of the target.

src/main/perl/Path.pm
jouvin: Not exactly. This is `$res` if the evaluated function succeeds. But it is also `undef` in case the evaluated function has a syntax or runtime error. in this case I agree that `$res` being undefined, the result should be the same.

src/main/perl/Path.pm
jouvin: I think this is the crucial point. `@_` is true only if there was a syntax or runtime error. But if `$res` is `undef` when the evaluated function completed but returned `undef` because of an error, then this is not catched by testing `@_ `. It took me quite some time to understand that is was the reason one of the test was failing.

src/main/perl/Path.pm
jouvin: Because I missed it! You mentioned it earlier pointing at `directory()` and this is where I found `_safe_eval()`. I thought it was what you were refering too. I'll double check if it makes things easier.

src/test/perl/path.t
jouvin: As said in https://github.com/quattor/CAF/pull/225#issuecomment-284447549, this is the plan! Just wanted to have the main things in place before tackling the details!

src/test/perl/path.t
jouvin: See https://github.com/quattor/CAF/pull/225#discussion_r104524627. Agreed.

src/main/perl/Path.pm
stdweird: you shouldn't use safe_eval on the LC::Check functions. LC::Check throws an error on failure and returns undef, but that's exactly what the `LC_Check` wrapper handles.
safe_eval is for "normal" perl functions.

i don't mind the code; but checking `$@` separately also makes sense, it can catch errors when functions return undef.

src/main/perl/Path.pm
stdweird: it's better in this case because it handles the exceptions thrown by LC, and respects the `keeps_state` option (the way you use `LC::Check` only uses global NoAction).

src/main/perl/Path.pm
jouvin: I understand now my mistakes and why I spent so much time to cope with the exceptions returned! Fixed!

src/main/perl/Path.pm
```diff
@@ -544,6 +565,103 @@ sub directory
...
+
+=item symlink
+
+Create a symlink C<link_path> whose target is C<target>.
+
+Returns undef and sets the fail attribute if C<link_path> 
+already exists and is not a symlink, except if this is a file and option C<force> is
+defined and true. If C<link_path> exists and is a symlink, it is updated.
+By default, the target is not required to exist. If you want to
+ensure that it exists, define option C<check> to true.
```
stdweird: add similar remark in hardlink staing that hardlink always checks, and that you cannot disable it (even if you pass `check=0`)

src/main/perl/Path.pm
```diff
@@ -544,6 +565,103 @@ sub directory
...
+This method relies on C<_make_link> method to do the real work,
+after enforcing the option saying that it is a symlink.
+
+=cut
+
+sub symlink
+{
+    my ($self, $target, $link_path, %opts) = @_;
+
+    $opts{hard} = 0;
```
stdweird: add comment like `# option passed to LC::Check::link`

src/main/perl/Path.pm
```diff
@@ -544,6 +565,103 @@ sub directory
...
+This method relies on C<_make_link> method to do the real work,
+after enforcing the option saying that it is a hardlink.
+
+=cut
+
+sub hardlink
+{
+    my ($self, $target, $link_path, %opts) = @_;
+
+    $opts{hard} = 1;
```
stdweird: add comment like `# option passed to LC::Check::link`

src/main/perl/Path.pm
```diff
@@ -544,6 +565,103 @@ sub directory
 }
 
 
+=item _make_link
+
+This method is mainly a wrapper over C<LC::Check::link>
+returning the standard CAF::Path return values. Every option
+supported by C<LC::Check::link> is supported. C<NoActionFlag>
```
stdweird: `C<NoActionFlag>` -> `C<NoAction>-flag`; also mention that it support the `keeps_state` option

src/main/perl/Path.pm
```diff
@@ -346,6 +352,21 @@ sub any_exists
...
+=item is_symlink
+
+Test if C<path> is a symlink.
+
+Returns true as long as C<path> is a symlink, even though the 
+symlink target doesn't exist.
+
+=cut
+
+sub is_symlink
```
stdweird: out of curiosity, why is there no `is_hardlink`?  maybe not `is_hardlink`, but rather `has_hardlink`.
i'm not saying we should have one, i'm just wondering.

src/main/perl/Path.pm
```diff
@@ -544,6 +565,103 @@ sub directory
...
+    }
+}
+
+=item hardlink
+
+Create a hardlink C<link_path> whose target is C<target>.
+
+On failure, returns undef and sets the fail attribute.
+If C<link_path> exists and is a file, it is updated. 
+C<target> must exist and must reside in the same file system
```
stdweird: `must exist` -> `must exist and cannot be a directory`. 
can you hardlink a symlink?

src/test/perl/path.t
```diff
@@ -290,44 +312,137 @@ rmtree if -d $basetest;
...
+
+my $target_directory = "tgtdir";
+my $target_file1 = "tgtfile1";
+my $target_file2 = "$basetest/$target_directory/tgtfile2";
+makefile("$basetest/$target_file1");
+makefile($target_file2);
+my %opts;
+is($mc->symlink($target_directory, $dirlink), CHANGED, "directory symlink created");
+ok($mc->is_symlink($dirlink), "$dirlink is a symlink");
+is(readlink($dirlink), $target_directory, "$dirlink has the expected target ($target_directory)");
```
stdweird: add a comment (once) that this tests more than it looks eg `# readlink returns undef on eg a regular file and a hardlink, so this truly tests that $dirlink is a symlink)

src/test/perl/path.t
```diff
@@ -290,44 +312,137 @@ rmtree if -d $basetest;
...
+
+init_exception("symlink tests");
+
+my $target_directory = "tgtdir";
+my $target_file1 = "tgtfile1";
+my $target_file2 = "$basetest/$target_directory/tgtfile2";
+makefile("$basetest/$target_file1");
+makefile($target_file2);
+my %opts;
+is($mc->symlink($target_directory, $dirlink), CHANGED, "directory symlink created");
```
stdweird: one missing loophole in this test: start with testing that `$dirlink` is not a symlink (if the symlink would already exist, this could still do not what you expect).

src/test/perl/path.t
```diff
@@ -290,44 +312,137 @@ rmtree if -d $basetest;
...
+ok($mc->is_symlink($dirlink), "$dirlink is a symlink");
+is(readlink($dirlink), $target_directory, "$dirlink has the expected target ($target_directory)");
+is($mc->symlink($target_file2, $filelink), CHANGED, "file symlink created");
+ok($mc->is_symlink($filelink), "$filelink is a symlink");
+is(readlink($filelink), $target_file2, "$filelink has the expected target ($target_file2)");
+is($mc->symlink($target_file2, $filelink), SUCCESS, "file symlink already exists: nothing done");
+ok($mc->is_symlink($filelink), "$filelink is still a symlink");
+is(readlink($filelink), $target_file2, "$filelink target not changed");
+is($mc->symlink($target_file1, $filelink), CHANGED, "file symlink already exists: nothing done");
```
stdweird: wrong message

src/test/perl/path.t
```diff
@@ -290,44 +312,137 @@ rmtree if -d $basetest;
...
+ok($mc->is_symlink($filelink), "$filelink is still a symlink (target changed)");
+is(readlink($filelink), $target_file1, "$filelink has the expected target ($target_file1)");
```
stdweird: `expected target` -> `expected changed target`

src/test/perl/path.t
```diff
@@ -290,44 +312,137 @@ rmtree if -d $basetest;
...
+$link_status = $mc->symlink($target_file1, "$test_directory", %opts);
+ok (! defined($link_status), "directory not replaced by a symlink (force option set)");
+is($mc->{fail},
+   "*** cannot symlink $test_directory: it is not an existing symlink",
+   "fail attribute set after symlink failure (existing directory not replaced by a symlink)");
+ok($mc->directory_exists($test_directory) && ! $mc->is_symlink($test_directory),
+   "Directory $test_directory has not be replaced by a symlink");
+$opts{check} = 1;
+ok(! $mc->symlink("really_really_missing", $brokenlink, %opts), "broken symlink not created (target existence enforced)");
+ok(! -e $brokenlink, "Broken link has not been created");
```
stdweird: this is not a test for a missing broken symlink, only `! -e $fn && ! -l $fn` is. this  one of the reason why broken symlinks are a real mess.

with broken symlink b
```
[stdweird@gast039e tmp]$ perl -e 'print -e "b", "X", -f "b", "X", -l "b", "\n"'
XX1
```

src/test/perl/path.t
```diff
@@ -290,44 +312,137 @@ rmtree if -d $basetest;
...
+delete $opts{check};
+is($mc->symlink("really_missing", $brokenlink), CHANGED, "broken symlink created");
+ok($mc->is_symlink($brokenlink), "Broken link has been created (target check disabled)");
+
```
stdweird: add a readlink test too

src/test/perl/path.t
```diff
@@ -290,44 +312,137 @@ rmtree if -d $basetest;
...
+ok($mc->_reset_exception_fail(), "_reset_exception_fail after symlink tests");
+
+# Test hardlink creation
+
+init_exception("hardlink tests");
+
+my $cwd = cwd();
+my $hardlink1 = "$basetest/$target_directory/hardlink1";
+my $hardlink2 = "$basetest/$target_directory/hardlink2";
+# hardlink target must be an absolute path or the link path directory is prepended
```
stdweird: add this feature/behaviour to the pod and add a test for passing a relative path

src/test/perl/path.t
```diff
@@ -290,44 +312,137 @@ rmtree if -d $basetest;
...
+$target_file1 = "$cwd/$basetest/tgtfile1";
+my $target_file3 = "$cwd/$basetest/$target_directory/tgtfile3";
+makefile($target_file1);
+makefile($target_file3);
+is($mc->hardlink($target_file1, $hardlink1), CHANGED, "hardlink created");
+is($mc->hardlink($target_file1, $hardlink1), SUCCESS, "hardlink already exists");
+is($mc->hardlink($target_file3, $hardlink1), CHANGED, "hardlink updated");
+$link_status = $mc->hardlink("missing_file", $hardlink1);
+ok(! $link_status, "hardlink not created (target doesn't exist)");
```
stdweird: `doesn't` -> `has to` (there's no check equivalent as with symlink. it really always has to exist

src/main/perl/Path.pm
jouvin: A hardlink is just another directory entry for an inode. There is no way to distinguish the "primary entry" from the new one (or the other ones). So there is no way to say that one particular file (directory entry) is a hardlink. `has_hardlink()` is probably feasable: just count the number of directory entries for an inode and if >1, means that the entry has hardlinks. Let's keep it for when we need it...

src/main/perl/Path.pm
jouvin: Yes you can! Only directories are not supported because not all file systems support hardlink to directories (see `man ln`, in fact `ln` need special options (`-d -F`) restricted to `root`). Example showing that it works for symlinks:
```
lx3/jouvin % echo "" > a_file
lx3/jouvin % ls -i a_file 
68 a_file
lx3/jouvin % ln -s a_file a_symlink
lx3/jouvin % ls -i a_symlink 
149 a_symlink
lx3/jouvin % ln a_symlink a_hardlink
lx3/jouvin % ls -i a_hardlink 
149 a_hardlink
```
You can see that `a_hardlink` has the same inode number as `a_symlink`(and is different from `a_file`).

src/test/perl/path.t
jouvin: Anyway there is the `is_symlink()` call just before... Do you suggest to remove it? (you are right that it is useless but it allows to give more details about the reason of the failure).

src/main/perl/Path.pm
stdweird: ok, thanks for trying this out.

src/test/perl/path.t
jouvin: But `$dirlink` is created as a symlink just before... This is like checking arguments in functions: do we assume that people writing/modifying unit tests know what they are doing and are not creating that is not a symlink for `$dirlink`?

src/test/perl/path.t
jouvin: As you prefer, means the same for me! (or `doesn't` does not suggest that you could override it).

jouvin:
Previous remarks addressed. Still missing tests for `NoAction`.

jouvin:
Tests should be complete now! At least ready for detailed scrutiny by @stdweird !

jouvin:
`has_hardlinks()` and `is_hardlink()` added as they were basically needed for the unit tests... PR should really be complete now!

src/main/perl/Path.pm
```diff
@@ -346,6 +353,21 @@ sub any_exists
     return $path && (-e $path || -l $path);
 }
 
+=item is_symlink
+
+Test if C<path> is a symlink.
+
+Returns true as long as C<path> is a symlink, even though the
```
stdweird: not a native speaker, but `though` -> `when` or `if`? @ned21 @jrha

src/main/perl/Path.pm
```diff
@@ -544,6 +566,178 @@ sub directory
 }
 
 
+=item _make_link
+
+This method is mainly a wrapper over C<LC::Check::link>
+returning the standard CAF::Path return values. Every option
```
stdweird: `CAF::Path` in `C<>`

src/main/perl/Path.pm
```diff
@@ -544,6 +566,178 @@ sub directory
...
+
+On failure, returns undef and sets the fail attribute.
+If C<link_path> exists and is a file, it is updated. 
+C<target> must exist (C<check> flag available in symlink()
+is ignored for hardlinks) and it must reside in the same 
+file system as C<link_path>. Also, if C<target_path> is a
+relative path, C<link_name> directory is prepended to it
+(C<link_name> can be a relative path): this way the API
+produces the same result for a symlink and a hardlink.
+C<link_name> directory is created if it doesn't exist.
```
stdweird: `directory` -> `parent directory`?

src/main/perl/Path.pm
jouvin: I'm interested by a lesson! I never know which one to use when!

src/main/perl/Path.pm
```diff
@@ -544,6 +566,178 @@ sub directory
...
+
+sub symlink
+{
+    my ($self, $target, $link_path, %opts) = @_;
+
+    # Option passed to LC::Check::link to indicate a symlink
+    $opts{hard} = 0;
+    
+    # LC::Check::symlink() expects an option 'nocheck' but CAF::Path::symlink exposes
+    # an attribute 'check', as the default in CAF::Path::symlink is not to check the
```
stdweird: `attribute` -> `option`

src/main/perl/Path.pm
```diff
@@ -544,6 +566,178 @@ sub directory
...
+{
+    my ($self, $file) = @_;
+    $file = $self->_untaint_path($file, "has_hardlinks") || return;
+    
+    if ( ! $self->file_exists($file) ) {
+        $self->debug(2, "has_hardlinks(): $file doesn't exist or is not a file");
+        return;
+    }
+    
+    my $nlinks = (stat($file))[3];
```
stdweird: this is tricky. since there can be hardlinks of a symlink, i think we should use `lstat` here (not to follow the link)

stdweird: also, i checked and couldn't find an immediate failure of stat (or lstat).
i would add a comment here that it is safe enough not to use `_safe_eval`

src/main/perl/Path.pm
```diff
@@ -544,6 +566,178 @@ sub directory
...
+    $self->debug(2, "Number of links to $file: $nlinks (hardlink if > 2)");
+    return $nlinks ? $nlinks - 1 : $nlinks;
```
stdweird: i would use `: 0` to make explicit it returns 0.

src/main/perl/Path.pm
```diff
@@ -544,6 +566,178 @@ sub directory
...
+}
+
+
+=item is_hardlink
+
+This method returns SUCCESS if C<link_path> and C<target_path> refer to the same file (inode).
+It returns 0 if C<link_path> and C<target_path> both exist but are different files and C<undef>
+if one of the paths doesn't exist or is not a file.
+
+Note: the result returned will be identical whatever is the order of C<link_path> and C<target_path>
```
stdweird: if this is the case, then i find the naming confusing. use `path1` and `path2` perhaps?

src/main/perl/Path.pm
```diff
@@ -544,6 +566,178 @@ sub directory
...
+    if ( ! $self->file_exists($link_path) ) {
+        $self->debug(2, "is_hardlink(): $link_path doesn't exist or is not a file");
+        return;
+    }
+    if ( ! $self->file_exists($target_path) ) {
+        $self->debug(2, "is_hardlink(): $target_path doesn't exist or is not a file");
+        return;
+    }
+    
+    my $link_inode = (stat($link_path))[1];
```
stdweird: same `lstat` remark as above.

src/test/perl/path.t
```diff
@@ -81,6 +82,11 @@ ok(! $mc->_get_noaction(1), "_get_noaction returns false with CAF::Object::NoAct
 
 my $exception_reset = 0;
 
+# init_exception() and verify_exception() functions work in pair. They allow to register a message
+# in 'fail' attribute at the beginning of a test section and to verify if new (unexpected) exceptions
+# where raised during the test section. To reset the 'fail' attribute after verify_exception(),
+# call _reset_exception_fail().  init_exception() implicitely reset the 'fail' attribute and also
```
stdweird: 2 spaces after `.`; `implicitely reset` -> `implicitly resets`

src/test/perl/path.t
stdweird: @jouvin this still needs to be addressed. i would revert the changes and keep the old logic. it's in any case more important that it fails rather that the message is readable/human friendly

src/test/perl/path.t
```diff
@@ -277,63 +298,242 @@ $mc->{fail} = undef;
 is($mc->_untaint_path("abc", "ok"), "abc", "untaint ok");
 ok(! defined($mc->{fail}), "no fail attribute set with ok path");
 
-=head2 directory/file/any exists
+
+=head2 directory/file/any exists on files and directoriees
```
stdweird: `directoriees` -> `directories`

src/test/perl/path.t
```diff
@@ -277,63 +298,242 @@ $mc->{fail} = undef;
...
 =cut
 
-init_exception("existence tests");
+init_exception("existence tests (file/directory)");
 
 # Tests without NoAction
 $CAF::Object::NoAction = 0;
 
-rmtree if -d $basetest;
+rmtree ($basetest) if -d $basetest;
```
stdweird: nice catch!

src/test/perl/path.t
```diff
@@ -277,63 +298,242 @@ $mc->{fail} = undef;
...
+=head2 symlink/hardlink creation/update/test
+
+=cut
+
+# Test symlink creation
+
+# Function to do the ok()/is() pair, taking into account NoAction flag
+sub check_symlink {
+    my ($mc, $target, $link_path, $expected_status) = @_;
+    my $noaction = $mc->_get_noaction();
```
stdweird: since you do not expect/support keeps_state, might as well use `CAF::Object::NoAction`

src/test/perl/path.t
```diff
@@ -277,63 +298,242 @@ $mc->{fail} = undef;
...
+    } else {
+        $ok_msg = "$link_path is " . ($expected_status == CHANGED ? "" : "already ") . "a symlink";
+    }
+    my $target_msg = "$link_path symlink has the expected " . ($expected_status == CHANGED ? "changed " : "") . "target ($target)";
+
+    # Test if symlink was created or not according to NoAction flag
+    my $ok_condition;
+    if ( $noaction ) {
+        $ok_condition = ! $mc->any_exists($link_path);
+    }else {
```
stdweird: space after `{`

src/main/perl/Path.pm
jouvin: I also checked `stat()` documentation and saw that in case of any failure it was returning undef which is what we need. I guess the main cause for failure is a file that doesn't exist but this is checked before... I'll change for `lstat()`, you are right that the case of hardlink of a symlink will be a problem with `stat()`. I'll add a unit test for this case.

src/test/perl/path.t
```diff
@@ -277,63 +298,242 @@ $mc->{fail} = undef;
...
+rmtree ($basetest) if -d $basetest;
+my $target_directory = "tgtdir";
+my $target_file1 = "tgtfile1";
+my $target_file2 = "$basetest/$target_directory/tgtfile2";
+makefile("$basetest/$target_file1");
+makefile($target_file2);
+my %opts;
+
+# Symlink creations
+for $CAF::Object::NoAction (1,0) {
```
stdweird: hah, never tought to do this! 👍

src/main/perl/Path.pm
jouvin: it was the intent! good catch!

src/test/perl/path.t
```diff
@@ -277,63 +298,242 @@ $mc->{fail} = undef;
...
+    is($mc->symlink($target_directory, $dirlink), CHANGED, "directory symlink created");
```
stdweird: add this to `check_symlink`?
but you should verify that the `link_path` does not exist prior to creating it

src/test/perl/path.t
jouvin: I have no problem supporting it but after thinking about it, I concluded that a specific unit test for this case was not needed as `keeps_state` is handled transparently by `_get_action()`. At the end there are 2 different cases: noaction=true and noaction=false, no matter how you reached this state. Adding a specific test for `keeps_state` would mean unit testing `_get_action()` not the link related methods... Or did I miss something?

src/test/perl/path.t
```diff
@@ -277,63 +298,242 @@ $mc->{fail} = undef;
...
-makefile("$basetest/tgtdir/tgtfile");
-ok(symlink("tgtdir", $dirlink), "directory symlink created");
-ok(symlink("tgtdir/tgtfile", $filelink), "file symlink created");
+# Valid symlink updates
+# The following tests make no sens if NoAction is true (require the previous creations to occur)
+is($mc->symlink($target_file2, $filelink), SUCCESS, "file symlink already exists: nothing done");
+check_symlink($mc, $target_file2, $filelink, SUCCESS);
+is($mc->symlink($target_file1, $filelink), CHANGED, "file symlink updated");
+check_symlink($mc, $target_file1, $filelink, CHANGED);
+is(readlink($filelink), $target_file1, "$filelink has the expected changed target ($target_file1)");
```
stdweird: is this different then thet readlink test in `check_symlink`?

src/test/perl/path.t
```diff
@@ -277,63 +298,242 @@ $mc->{fail} = undef;
...
+my $link_status = $mc->symlink($target_file1, $target_file2);
+ok(! defined($link_status), "symlink failed: existing file not replaced by a symlink");
+is($mc->{fail},
+   "*** cannot symlink $target_file2: it is not an existing symlink",
+   "fail attribute set after symlink failure (existing file not replaced by a symlink)");
+ok($mc->file_exists($target_file2) && ! $mc->is_symlink($target_file2), "File $target_file2 has not be replaced by a symlink");
+$opts{force} = 1;
+$CAF::Object::NoAction = 1;
+is($mc->symlink($target_file1, $target_file2, %opts), CHANGED, "existing file replaced by a symlink (force option set)");
+ok($mc->file_exists($target_file2), "File $target_file2 not replaced by a symlink with 'force' option (NoAction set)");
```
stdweird: hmm, if target_file2 would be a symlink, this test would pass. maybe do a `ok(!readlink($target_file2)`

src/test/perl/path.t
```diff
@@ -277,63 +298,242 @@ $mc->{fail} = undef;
...
+ok($mc->directory_exists($test_directory) && ! $mc->is_symlink($test_directory),
+   "Directory $test_directory has not be replaced by a symlink");
+
+# Broken symlinks with and without 'check' option
+$opts{check} = 1;
+ok(! $mc->symlink("really_really_missing", $brokenlink, %opts), "broken symlink not created (target existence enforced)");
+ok(! -e $brokenlink && ! -l $brokenlink, "Broken link has not been created");
+delete $opts{check};
+is($mc->symlink("really_missing", $brokenlink), CHANGED, "broken symlink created");
+ok($mc->is_symlink($brokenlink), "Broken link has been created (target check disabled)");
```
stdweird: check is not disabled, this is the default. you should add a test that has check disabled (`$opts{check} = 0`)

src/test/perl/path.t
stdweird: it's probably not due to removal of the base directory, but i can't seen what `dirlink` is.

src/test/perl/path.t
```diff
@@ -277,63 +298,242 @@ $mc->{fail} = undef;
...
+ok(! $link_status, "hardlink not created (do not replace an existing directory)");
+is($mc->{fail},
+   "*** cannot hard link $basetest/$target_directory: it is a directory",
+   "fail attribute set after hardlink error (existing directory not replaced)");
+$link_status = $mc->hardlink($relative_target_file3, $hardlink2);
+ok(! $link_status, "hardlink not created (relative target, link path directory prepended)");
+is($mc->{fail},
+   "*** invalid target ($relative_target_file3): stat(". dirname($hardlink2) . "/$relative_target_file3): No such file or directory",
+   "fail attribute set after hardlink error (relative target, link path directory prepended)");
+
```
stdweird: no unittests for has_hardlink / is_hardlink

src/test/perl/path.t
stdweird: no need, i think `get_action` is unittested. i was merely pointing out you could just as well use `CAF::Object::NoAction` directly

src/test/perl/path.t
jouvin: I thought about this but I think that `path.t` is easier to read with a separate creation of the link and check. In addition, it allow to use `check_symlink()` with slightly different creation scheme.

src/test/perl/path.t
jouvin: leftover...

src/test/perl/path.t
jouvin: Always the same remark... In principle yes... but the test writer should know what it is doing... Anyway easy to add...

src/test/perl/path.t
jouvin: `check=0` and `check` undefined at processed the same way... Anyway harmless to test both case...

src/test/perl/path.t
jouvin: As stated in the comment, `check_hardlink()` is the main unit test for these methods. I don't see the point of adding something specific that will do exactly the same thing (we need to create something and check that it works)... I'll also add these for cases where the argument is not a hardlink in the previous section of tests.

jouvin:
@ned21 @jrha @stdweird Adding unit tests for hardlinks to symlinks, which is allowed by `ln` command whatever the symlink target is, I found a "bug" in `LC::Check::symlink`. It explicitly disables this if the symlink target doesn't exist or is a directory, to be compatible with `ln` according to the comment! Probably an old version of `ln` was preventing this...
I have a fix for this but I don't know if we want to patch `LC::Check`. Without this fix the current PR cannot validate the test but if we decide not to patch `LC::Check`, I can easily remove the failing tests... What do you think?
BTW, `LC::Check` has another bug with respect to prepending the link parent directory to the target if the target is a relative path. It does it to test the existence (which is sensible for a symlink as this is the way it will be interpreted) but which is a bug in the case of hardlink as the test if not against what will be created... Thus the confusing comment that I put in `CAF::Path::hardlink` which is wrong in fact, as long as the result is concerned (but true as long as the existence check is concerned...). Another thing that probably need fixing...
What do you think?

jouvin:
PR should be almost ok but I still have to figure out the problem with `verify_exceptions()`. Role of the `noreset` argument is unclear for me. @stdweird could you give me some info?

jouvin:
I realized that we had a test-specific version of `LC::Check` (and `Process` and `Files`). It caused some tests to fail because the version of the module was not advertized but `CAF::Path` had a requirement on it. It is unclear for me why this was happening for certains tests only (seems to be those for modules using `CAF::Path`, except `CAF::Path` itself). Do we really need these test-specific versions of LC modules. Just removing them is unfortunately not working, so there are probably some subtilties...

jouvin:
BTW, there is still one test failing in `path.t` (that was not failing before) and I cannot reproduce it on my local machine... Strange...

jouvin:
Ready for final review 😄

src/main/perl/Path.pm
ned21: "even though" would normally be used when dealing with a non-conditional event. e.g. "The politician decided to vote against the law even though his or her party was for it." Whereas in this case it's not a given that the symlink does not exist so if or when would be more appropriate.

```Returns true as long as C<path> is a symlink, even if the symlink target does not exist.``` or
```Returns true as long as C<path> is a symlink, including when the symlink target does not exist.```

File: src/main/perl/Path.pm [hunk 65]
```diff
@@ -544,6 +566,180 @@ sub directory
...
+=item _make_link
+
+This method is mainly a wrapper over C<LC::Check::link>
+returning the standard C<CAF::Path> return values. Every option
+supported by C<LC::Check::link> is supported. C<NoAction>
+flag is handled by C<LC::Check::link> and C<keeps_state> option
+is honored (overrides C<NoAction> if true). One important
+difference is the order of the arguments: C<CAF::Path:_make_link>
+and the methods based on it are following the Perl C<symlink>
+(and Shell C<ln>) argument order.
```
stdweird: `ln` has nothing to do with `Shell`, it's part of base utils on a lot of Unix systems.

src/main/perl/Path.pm
```diff
@@ -544,6 +566,180 @@ sub directory
...
+
+This in internal method, not supposed to be called directly.
```
stdweird: `in` -> `is an`

src/main/perl/Path.pm
```diff
@@ -544,6 +566,180 @@ sub directory
...
+
+=item hardlink
+
+Create a hardlink C<link_path> whose target is C<target>.
+
+On failure, returns undef and sets the fail attribute.
+If C<link_path> exists and is a file, it is updated. 
+C<target> must exist (C<check> flag available in symlink()
+is ignored for hardlinks) and it must reside in the same 
+file system as C<link_path>. If C<target_path> is a
```
stdweird: `file system` -> `filesystem`

src/test/perl/path.t
stdweird: @jouvin ping

src/test/perl/path.t
```diff
@@ -277,63 +310,271 @@ $mc->{fail} = undef;
...
+ok(! $mc->is_symlink($basetest), "is_symlink false on created directory");
+ok(! $mc->has_hardlinks($basetest), "has_hardlinks false on created directory");
+is($mc->is_hardlink($basetest, $basetest), undef, "is_hardlink false with created directories");
 
 ok(! $mc->directory_exists($basetestfile), "directory_exists false on created file");
 ok($mc->any_exists($basetestfile), "any_exists true on created file");
 ok($mc->file_exists($basetestfile), "file_exists true on created file");
+ok(! $mc->is_symlink($basetestfile), "is_symlink false on created file");
+ok(! $mc->has_hardlinks($basetestfile), "has_hardlinks false on created file");
+is($mc->is_hardlink($basetestfile, $basetestfile), 0, "is_hardlink false with a non-hardlinked file");
```
stdweird: this is false because the files are the same. should change the message

src/test/perl/path.t
```diff
@@ -277,63 +310,271 @@ $mc->{fail} = undef;
...
+ok(! readlink($target_file2), "$target_file2 exists and is a file");
+my $link_status = $mc->symlink($target_file1, $target_file2);
+ok(! defined($link_status), "symlink failed: existing file not replaced by a symlink");
+is($mc->{fail},
+   "*** cannot symlink $target_file2: it is not an existing symlink",
+   "fail attribute set after symlink failure (existing file not replaced by a symlink)");
+ok($mc->file_exists($target_file2) && ! $mc->is_symlink($target_file2), "File $target_file2 has not be replaced by a symlink");
+$opts{force} = 1;
+$CAF::Object::NoAction = 1;
+is($mc->symlink($target_file1, $target_file2, %opts), CHANGED, "existing file replaced by a symlink (force option set)");
```
stdweird: adapt message. this is force=1, but also NoAction=1; so the file shouldn't be replaced.

src/test/perl/path.t
stdweird: please do add. this is testing NoAction behaviour

stdweird:
@jouvin more minor fixes. can you address them in a separate commit(s)? you can squash them before merging (it would make the review easier)

PR #228, opened 2017-03-23
CAF: fix pod issues for rebuild to rst.

wdpypere:
It seems rst (and pod2rst) is a bit stricter than md. This:

 - fixes several empty ```=item``` (which makes not a lot of sense anyway.)
 - newlines before and after code blocks

I don't know if this is OK or if you would rather have me split this up in several PR's.

### DIFF
--- a/src/main/perl/Object.pm
+++ b/src/main/perl/Object.pm
@@ -180,11 +180,13 @@ from the C<ENV> attribute hashref.
 Returns the C<env> hashref.
 
 To be used as
+
     # Setup local environment
     local %ENV = %ENV;
     $self->update_env(\%ENV);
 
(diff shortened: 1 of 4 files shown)


stdweird:
is there a way we can try to the conversion, so we can add it to 000-tqu doc tests?

wdpypere:
Yes. In short we could do something like:
```
pod2rst --infile CAF/target/doc/pod/CAF/Object.pod  --outfile=/tmp/test/test.rst
sphinx-build -C /tmp/test2 /tmp/test3 | egrep "ERROR|WARNING|SEVERE"
```

sphinx-build is so kind to always exit with 0....

PR #229, opened 2017-03-24
Path: move: do not cleanup dest, but add backup of dest using hardlink

stdweird:
Fixes https://github.com/quattor/configuration-modules-core/issues/991

### DIFF
--- a/src/main/perl/Path.pm
+++ b/src/main/perl/Path.pm
@@ -819,10 +819,15 @@ sub move
 
     return SUCCESS if (! $self->any_exists($src));
 
-    # Cleanup dest, use backup
-    if (! $self->cleanup($dest, $backup, %opts)) {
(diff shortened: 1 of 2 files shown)


ned21:
LGTM. Any chance of splitting the whitespace and comment cleanups earlier in the file into a separate commit?

stdweird:
aaaarghhhhh. that's my save.
agree to enforce http://search.cpan.org/~thaljef/Perl-Critic/lib/Perl/Critic/Policy/CodeLayout/ProhibitTrailingWhitespace.pm?

ned21:
@stdweird I'm fine with the policy in principle (it's same as pep8, right?) but a little bit nervous about the amount of clean up required?

stdweird:
@ned21 should be trivial in any editor; and most components are not on `00-tqu` yet.

stdweird:
@ned21 whitespace changes in separate commit

PR #230, opened 2017-03-26
Path: listdir method to return directory entry names

stdweird:
Based on #243

### DIFF
--- a/src/main/perl/FileWriter.pm
+++ b/src/main/perl/FileWriter.pm
@@ -306,8 +306,10 @@ sub close
                 my $parent_dir = dirname($filename);
                 # Pass NoAction here, as it keeps track of the NoAction value during initialisation and/or keeps_state
                 my $cafpath = CAF::Path::mkcafpath(log => *$self->{LOG}, NoAction => $options->{noaction});
-                # only create the directory if it didn't exist yet
-                # if not, this would change the directory mode on existing directories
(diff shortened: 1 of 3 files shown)


stdweird:
Main reason to wrap this is to able to unittest readdir in maven-tools with the mocked files

src/main/perl/FileWriter.pm
```diff
@@ -301,10 +303,21 @@ sub close
                 $msg = 'would have been';
                 $self->debug(1, "File $filename with NoAction=1");
             } else {
+                my $parent_dir = dirname($filename);
+                # Pass NoAction here, as it keeps track of the NoAction value during initialisation and/or keeps_state
+                my $cafpath = CAF::Path::mkcafpath(log => *$self->{LOG}, NoAction => $options->{noaction});
+                # only create the directory if it didn't exist yet
+                # if not, this would for the directory mode on existing directories
```
ned21: This line doesn't make sense. "this would [what?] for the"

stdweird:
@ned21 i'll address the remark, but i made a mistake in the description. this is based on the already merged PR #243; and your remark is from that PR. i'll add a new commit with better description

ned21:
@stdweird I thought I had seen that odd sentence before, and assumed I had just reviewed this PR before!

stdweird:
@ned21 i added a new commit to this PR

stdweird:
retest this please

PR #231, opened 2017-03-29
Update build tools to 1.53

jouvin:
- Fix resulting problems


jouvin:
As it has been merged already, I changed the milestone to 17.3 but @jrha feel free to change it back to 17.4.

PR #238, opened 2017-04-03
FileWriter/Editor: fix bug in close with source

### LINKED ISSUE #236: FileWriter: confusing same message with `cancel` and `close` with save=0
both cases report same message with verbose level

### LINKED ISSUE #237: CAF::FileEditor from source broken
The diff from FileEditor with source is broken; regular close for FileEditor works by accident (it rereads the original contents). All thanks to typo.

### LINKED ISSUE #239: CAF::FileWriter: reported diff does not end up in verbose logfile
Could be an issue with `CAF::Reporter`

stdweird:
Fixes #237 
Also fixes #236 and fixes #239

### DIFF
--- a/src/main/perl/FileWriter.pm
+++ b/src/main/perl/FileWriter.pm
@@ -229,21 +229,29 @@ sub close
 
     if (*$self->{save}) {
         *$self->{save} = 0;
-        my $content_ref = *$self->{buf};
+        my $content_ref = $self->string_ref();
+
+        my $report_diff = sub {
+            my ($diff, $msg) = @_;
+            if(*$self->{options}->{sensitive}) {
+                $self->verbose("Changes $msg $filename are not reported due to sensitive content");
+            } else {
+                $self->verbose("Changes $msg $filename:");
+                if ($self->is_verbose()) {
+                    $self->report($diff);
+                } elsif ($self->is_verbose(verbose_logfile => 1)) {
+                    $self->log($diff);
+                }
+            }
+        };
 
         if (*$self->{original_from_source}) {
             # Report changes compared to source
             my $src_original_content = *$self->{original_content};
             if (defined($src_original_content)) {
                 my $src_diff = diff(\$src_original_content, $content_ref, { STYLE => "Unified" });
-                my $src_changed = $src_diff ? 1 : 0;
-                if ($src_changed) {
-                    if(*$self->{options}->{sensitive}) {
-                        $self->verbose("Changes compared to source for $filename are not reported due to sensitive content");
-                    } else {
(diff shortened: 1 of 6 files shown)


stdweird:
i could squash the `905` and `ee7` commits; the `ee7` one makes the problem clear and was what the old code did.

src/main/perl/FileWriter.pm
```diff
@@ -215,21 +231,43 @@ sub close
...
+                $self->verbose("No original source content for $filename");
+            }
+        }
+
         # Always read and (try to) determine the diff
-        my $content_orig;
-        if (defined(*$self->{content_orig})) {
-            $self->debug(2, "Using original contents from $filename");
-            $content_orig = *$self->{content_orig};
+        my $original_content;
```
ned21: would it not be clearer to make this an elsif clause of the previous block and drop the ```!*$self->{original_from_source}``` test?

src/main/perl/FileWriter.pm
```diff
@@ -215,21 +231,43 @@ sub close
...
+        if (defined(*$self->{original_content}) && !*$self->{original_from_source}) {
+            $self->debug(2, "Using existing original content for $filename");
+            $original_content = *$self->{original_content};
         } else {
             # missing_ok=1 mimics original LC::Check::file behaviour
-            $content_orig = $self->_read_contents($filename, event => \%event, missing_ok => 1);
+            $original_content = $self->_read_contents($filename, event => \%event, missing_ok => 1);
```
ned21: Why is this extra read needed here? The conditions under which this line is executed are not trivially fathomable (cf. comment above) so I think some extra comments here about how we can end up here and why we need this line would be useful.

src/main/perl/FileWriter.pm
stdweird: hmmm, i'd rather not. the previous block is only about reporting, this block is about setting the content to use to determine change. i don't want to mix these, this is the critical bit of the close method.
maybe the attribute names can be changed?
i can also add that the first block is the "normal" path for FileEditor w/o source; the second path is the normal for FileWriter and FileEditor w source.
i just added the role and meaning of the attributes to the pod to make it clear what happens, i don't assume people should use them directly. maybe i should move the last block in the pod to regular comment?

src/main/perl/FileWriter.pm
stdweird: it's not extra for regular FileWriter (this is the first and only time the original file is read) and regular FIleEditor (ie without source; that is the other block) use.

stdweird:
@ned21 please rereview

ned21:
Thanks for adding the extra comments to clarify that if block.

PR #240, opened 2017-04-05
FileWriter/Editor: add reopen method

stdweird:
Requires #238

### DIFF
--- a/src/main/perl/FileEditor.pm
+++ b/src/main/perl/FileEditor.pm
@@ -143,6 +143,19 @@ no warnings 'redefine';
 *open = \&new;
 use warnings;
 
+# Open FileEditor starts at the end, so same here
+# Original content is now from current content, not from source
+sub reopen
+{
+    my $self = shift;
+    $self->SUPER::reopen();
+
+    # pass a copy
+    my $txt = *$self->{original_content};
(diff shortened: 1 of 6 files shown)


jrha:
LGTM, but I'm no perl/CAF guru.

ned21:
Looks like only 37db65a is new and the rest already merged so this needs a rebase.

Regarding 37db65a, I am not sure I understand why this is useful and where it is intended to be used. I'm unsure as to whether that is best made part of the POD for the method, or in the commit log.

stdweird:
@ned21 i wrote this code because i was trying to do something clever in the network rewrite with initial write and close from source, and then another write (in summary: i was too lazy of call copy 😄 ).

i now also see 2 minor changes that could be useful otherwise: handle close an already closed file and handle the stringification of a closed file.

ned21:
tbh I'm not convinced this is useful. It's an extra function which has to be maintained and tested against all future changes -- additional rarely used code is a maintenance burden -- and does it really save any code for the user?

stdweird:
@ned21 ok. i'll try to extract the 2 minor changes/fixes that are useful in another PR. we can always reopen this PR is guess

PR #246, opened 2017-04-26
bump build-tools to 1.54

stdweird:


stdweird:
not stricly needed for 17.3, but better put all repos on same baseline

PR #247, opened 2017-06-19
remove set_caf_file_close_diff and unneeded NoAction from rbe unittests

stdweird:


stdweird:
retest this please

ned21:
Could the commit log explain why are they unneeded/safe to remove please?

stdweird:
@ned21 not how to formulate that
* `set_caf_file_close_diff` is a noop in 1.54 and scheduled for removal in 1.55 (https://github.com/quattor/maven-tools/pull/162)
* the rbe unittests do not test NoAction, so it is not needed to specify that anymore (NoAction was used in unittests to prevent actual files being written etc; but since 1.54, this is no longer needed as everything is mocked when using `Test::Quattor` in unittests; and setting NoAction should only be done to test NoAction behaviour)

ned21:
@stdweird What you have written in the comment is perfect - it just needs to be in the commit log so we can preserve it if GH disappears / we migrate to something else.

stdweird:
@ned21 commit message modified

jrha:
@hpcugentbot test this please

PR #248, opened 2017-07-26
Path: clarify how listdir method can replace glob()

stdweird:


stdweird:
@jrha looks familiar? 😄

PR #250, opened 2017-08-08
FileWriter: DESTROY: do not close an already closed file

stdweird:
Fixes the issue where already/explicitly closed FileWriter instances get re-closed during object destroy, causing a unnecessary verbose message that is very confusing.


stdweird:
@jrha better?

PR #256, opened 2017-11-09
Process: add sensitive option to prevent verbose reporting of the command

### LINKED ISSUE #254: CAF::Process: Include sensitive option
Sometimes you need to execute a command that includes a password for example. This is quite sensitive and we should avoid to show the command from the logs. 

We can avoid this if we include an `sensitive`   option within ` %opts`    in these cases. If the option is set then CAF::Process does not show the sensitive command from the log.

stdweird:
Fixes #254

### DIFF
--- a/src/main/perl/Process.pm
+++ b/src/main/perl/Process.pm
@@ -103,6 +103,17 @@ regardless of any value for C<NoAction>.
 By default, commands modify the state and thus C<keeps_state> is
 false.
 
+=item C<sensitive>
+
+A boolean specifying whether the arguments contain sensitive information
+(like passwords). If C<sensitive> is true, the commandline will not be reported
+(by default when C<log> option is used, the commandline is reported
+with verbose level).
+
+This does not cover command output. If the output (stdout and/or stderr) contains
+sensitve information, make sure to handle it yourself via C<stdout> and/or C<stderr>
+options (or by using the C<output> method).
+
 =back
 
 These options will only be used by the execute method.
@@ -121,6 +132,8 @@ sub _initialize
         $self->debug(1, "keeps_state set");
         $self->{NoAction} = 0
     };
+
+    $self->{sensitive} = $opts{sensitive};
 
     $self->{COMMAND} = $command;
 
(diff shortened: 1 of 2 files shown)


src/main/perl/Process.pm
```diff
@@ -103,6 +103,16 @@ regardless of any value for C<NoAction>.
 By default, commands modify the state and thus C<keeps_state> is
 false.
 
+=item C<sensitive>
+
+A boolean specifying whether the command (and args) contain sensitive information
+(like passwords). If C<sensitive> is true, the commandline will not be reported
+(by default the commanlione is reported in with verbose).
```
ned21: commanlione => command line. Also "by default" do you mean only if the `log` option is not `undef`? Or do you really mean "in all cases unless you set this" ?

src/main/perl/Process.pm
```diff
@@ -103,6 +103,16 @@ regardless of any value for C<NoAction>.
...
+
+This does not cover command output. If th eoutput (stdout or stderr) contains
```
ned21: the output

src/main/perl/Process.pm
```diff
@@ -103,6 +103,16 @@ regardless of any value for C<NoAction>.
...
+sensitve information, make sure to handle it yourself via C<stdout> and/or C<stderr>
```
ned21: Should reference the `log` option here? (i.e. recommend setting it to undef?) Generally I am a bit confused as to how this interacts with the log option so it would be good to see that clarified.

stdweird: i'm not sure how to formulate it. the code here does not report the output, but eg `execute` sends it to the real stdout (and that's also why it's not part of the logs, see #245).

stdweird:
@ned21 i made some changes to address your remarks, but also changed that the sensitive option only applies ot the arguments. the binary itself can be reported by some other methods also (and i'm not sure the binary itself is ever sensitive)

src/main/perl/Process.pm
```diff
@@ -103,6 +103,17 @@ regardless of any value for C<NoAction>.
 By default, commands modify the state and thus C<keeps_state> is
 false.
 
+=item C<sensitive>
+
+A boolean specifying whether the arguments contain sensitive information
+(like passwords). If C<sensitive> is true, the commandline will not be reported
+(by default and when C<log> option is used, the commandline is reported
```
ned21: Sorry to nitpick but I think "by default when the C<log> option is used" is clearer.

ned21:
Thanks for the changes. All good except one minor nitpick about removing what I believe to be an extraneous "and". Then you can squash.

stdweird:
@ned21 should be fine now

jrha:
@ned21 are you happy with this now?

PR #257, opened 2017-11-15
Process: run process as another user

### LINKED ISSUE #253: CAF::Process: Include an option to execute the command as an specific user
It would be nice to include a new option within `%opt`    to execute the command as an specific user instead of root. 
Maybe `$opts{user}`    for example.

stdweird:
Fixes #253

### DIFF
--- a/src/main/perl/Process.pm
+++ b/src/main/perl/Process.pm
@@ -2,7 +2,7 @@
 
 use parent qw(CAF::Object);
 
-use LC::Exception qw (SUCCESS throw_error);
+use LC::Exception qw (SUCCESS);
 use LC::Process;
 
 use File::Which;
@@ -11,6 +11,8 @@ use File::Basename;
 use overload ('""' => 'stringify_command');
 use Readonly;
 
+use English;
+
 Readonly::Hash my %LC_PROCESS_DISPATCH => {
     output => \&LC::Process::output,
     toutput => \&LC::Process::toutput,
@@ -18,7 +20,6 @@ Readonly::Hash my %LC_PROCESS_DISPATCH => {
     trun => \&LC::Process::trun,
     execute => \&LC::Process::execute,
 };
-
 
 =pod
 
@@ -138,6 +139,27 @@ This does not cover command output. If the output (stdout and/or stderr) contain
 sensitve information, make sure to handle it yourself via C<stdout> and/or C<stderr>
 options (or by using the C<output> method).
 
+=item C<user>
+
+Run command as effective user. The C<user> can be an id (all digits) or a name.
+
+This only works when the current user is root.
+
+In case a non-root user uses this option, or C<user> is not a valid user,
+the initialisation will work but any actual execution will fail.
+
+=item C<group>
+
+Run command with effective group. The C<group> can be an id (all digits) or a name.
+
+If C<user> is defined, and C<group> is not, the users primary group will be used
+(instead of the default root group).
+
+This only works when the current user is root.
+
+In case a non-root user uses this option, or C<group> is not a valid group,
+the initialisation will work but any actual execution will fail.
+
 =back
 
 These options will only be used by the execute method.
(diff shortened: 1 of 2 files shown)


stdweird:
This PR needs unittests, but i'm still thinking how to do them.
But the code was tested and works.

src/main/perl/Process.pm
```diff
@@ -153,10 +177,125 @@ C<<<msg> command: <COMMAND>[ <postmsg>]>>.
 
 =cut
 
+sub _get_uid_gid
+{
+    my ($self, $mode) = @_;
+
+    my @res;
+    my $what = $self->{$mode};
```
ned21: I find $what to be a bit vague. How about $targetid ?

src/main/perl/Process.pm
```diff
@@ -165,16 +304,26 @@ sub _LC_Process
...
-        my $funcref = $LC_PROCESS_DISPATCH{$function};
-        if (defined($funcref)) {
-            return $funcref->(@$args);
-        } else {
-            $self->error("Unsupported LC::Process function $function");
-            return;
+        # The original GID (as list of groups)
+        my $orig_user_group = [$UID, "$GID"];
+
+        if ($self->_set_eff_user_group()) {
```
ned21: I think my main reservation about this PR is that setting the user/group is going to be a rare code path, the common case will remain that user/group are unset and things run as root. So that means for someone try to understand the code they now have to read  three extra levels of function depth to confirm that it does nothing.

So while it's redundant to do so, I think from a code maintenance point of view it would be better to gate the call to _set_eff_user_group() on user/group being set.

stdweird: i understand.

so you mean something like 
```perl
my $current_user_group = (!defined $self->{user}) or (!defined $self->{group});
if ($current_user_group or $self->_set_eff_user_group()) {
 ...
}
```
and an extra conditional `if ! $current_user_group` for the restore part?

src/main/perl/Process.pm
stdweird: i'll use `target` (the id part suggests only ids are used, while i think most likely people will use usernames instead of ids.

stdweird:
@ned21 please rereview. tests fail, will be fixed later

stdweird:
tests are complete

jrha:
@hpcugentbot test this please

jrha (dismissed):

stdweird:
@jrha i'm moving the milestone. we are seeing something odd, and at least the reporting needs refinement.

jrha:
@stdweird any update on the strange behaviour?

jrha:
@stdweird any update? Still seeing strangeness or is this now understood and ready for merging?

stdweird:
@jrha i didn't find the time yet to debug further. i moved the milestone.

jrha:
:+1:

jrha:
@stdweird got any further?

stdweird:
not really. sometimes, for whatever reason, the current GID is a single group; and in that case something goes wrong try to restore the group (looks like you can reduce th enumebr of effective groups one is in)
eg. here the starting GID is `0` (normally it's a more than one group), and the resulting group 
```
2018/06/05-10:24:23 [VERB] Got user id 163 (gid 163) (is_id 0 is_user 1)
2018/06/05-10:24:23 [VERB] Changing EGID from '0' to '163 163' with GID '0'
2018/06/05-10:24:23 [VERB] Changing EUID from '0' to 163 with UID 0
2018/06/05-10:24:25 [VERB] Restoring EUID from '163' to 0 with UID 0
2018/06/05-10:24:25 [ERROR] Something went wrong restoring EGID from '163 163' to '0' with GID '0 163': new EGID '0 163', no reason given
```

jrha:
Hmm, odd.

jrha:
Bumping this again, will see if I can take some time to look it more closely.

jrha:
Rebased.

jrha:
Rebased on `main`

PR #262, opened 2017-12-01
FileWriter: set file status when no content was modified

### LINKED ISSUE #259: Filewriter: modify permissions when content is unmodified
Currently when the content is unmodified, other changes like owner/mode are not forced to the file.
I propose to fix this and report any changes to owner and/or mode also as a changed file on close.

stdweird:
Fixes #259

### DIFF
--- a/src/main/perl/FileWriter.pm
+++ b/src/main/perl/FileWriter.pm
@@ -7,7 +7,7 @@ use File::AtomicWrite 1.18;
 use Errno qw(ENOENT);
 use IO::String;
 use CAF::Process;
-use CAF::Object;
+use CAF::Object qw(CHANGED);
 use CAF::Path;
 use File::Basename qw(dirname);
 use overload '""' => "stringify";
@@ -298,6 +298,9 @@ sub close
 
         my $msg = 'was';
 
+        # Pass NoAction here, as it keeps track of the NoAction value during initialisation and/or keeps_state
+        my $cafpath = CAF::Path::mkcafpath(log => *$self->{LOG}, NoAction => $options->{noaction});
+
         if ($changed) {
             $report_diff->($diff, 'to');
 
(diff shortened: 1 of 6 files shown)


stdweird:
@jrha i tagged this for 17.12 (we need it for some security related changes we made to some of our configs, but if this is too late, feel free to set it to 18.3 again)

src/main/perl/FileWriter.pm
```diff
@@ -346,7 +347,40 @@ sub close
...
             }
         } else {
-            $msg = 'was not';
+
+            # even when nothing changed set the permissions
+            my %fileopts;
+            foreach my $name (qw(mode mtime owner group)) {
+                $fileopts{$name} = $options->{$name} if exists($options->{$name});
+            };
+            # probably alwasy true since mode has a default value
```
jrha: alwasy →always

stdweird:
@jrha typo fixed

src/test/perl/filewriter-notmocked.t
```diff
@@ -80,6 +80,15 @@ is(readfile($fn), $TEXT, "file $fn created with expected contents");
 is(get_perm($fn), 0764, "created file $fn has mode 0764");
 is((stat($fn))[9], 1234567, "mtime set to 1234567 forfile $fn");
 
+$fh = CAF::FileWriter->open ($fn, mode => 0764, log => $obj);
+print $fh $TEXT;
+ok(!$fh->close(), 'close reports no change w same content and same permissions');
+
+$fh = CAF::FileWriter->open ($fn, mode => 0700, log => $obj);
+print $fh $TEXT;
+ok($fh->close(), 'close reports change w same content and different permissions');
```
ned21: I don't think there's any reason to keep lines short is there? In which case it would be better to have the full "with" instead of just w which is a bit of an idiom and might be harder for non-experts in English to comprehend?

src/test/perl/filewriter-notmocked.t
```diff
@@ -80,6 +80,15 @@ is(readfile($fn), $TEXT, "file $fn created with expected contents");
...
+is(get_perm($fn), 0700, "created file $fn has mode 0700 w same content");
```
ned21: w => with

src/test/perl/filewriter-notmocked.t
```diff
@@ -80,6 +80,15 @@ is(readfile($fn), $TEXT, "file $fn created with expected contents");
 is(get_perm($fn), 0764, "created file $fn has mode 0764");
 is((stat($fn))[9], 1234567, "mtime set to 1234567 forfile $fn");
 
+$fh = CAF::FileWriter->open ($fn, mode => 0764, log => $obj);
+print $fh $TEXT;
+ok(!$fh->close(), 'close reports no change w same content and same permissions');
```
ned21: w => with

stdweird:
@ned21 just me being lazy. i replaced the ` w `'s

PR #265, opened 2018-04-26
TextRender: add jsonpretty module

kwaegema:
add module to Indent json and add spaces before and after

### DIFF
--- a/src/main/perl/TextRender.pm
+++ b/src/main/perl/TextRender.pm
@@ -422,10 +422,17 @@ sub load_module
     return 1;
 }
 
-
-sub render_json
+# Indent json and add spaces before and after
+sub render_jsonpretty
 {
     my ($self) = @_;
+    return $self->render_json(1);
+}
+
+# render json output, if pretty is true, this is rendered with idents
+sub render_json
+{
+    my ($self, $pretty) = @_;
 
     # We should only support hash or array refs (see JSON::XS allow_nonref option)
     # JSON::XS croaks if not handled properly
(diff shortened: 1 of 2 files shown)


src/main/perl/TextRender.pm
```diff
@@ -422,17 +422,24 @@ sub load_module
...
+sub render_jsonpretty
+{
+    my ($self) = @_;
+    return $self->render_json(1);
+}
 
 sub render_json
 {
-    my ($self) = @_;
+    my ($self, $pretty) = @_;
```
stdweird: add line of comment what this method and the optional arg does

src/main/perl/TextRender.pm
```diff
@@ -422,17 +422,24 @@ sub load_module
...
 
     # We should only support hash or array refs (see JSON::XS allow_nonref option)
     # JSON::XS croaks if not handled properly
     my $ref = ref($self->{contents});
     if ($ref eq 'HASH' || $ref eq 'ARRAY') {
         my $j = JSON::XS->new();
         $j->canonical(1); # sort the keys, to create reproducable results
+        $j->pretty(1) if (defined($pretty) && $pretty);
```
stdweird: just use `if !$pretty`, no need for `defined` as `undef` evaluates to false

stdweird: it might even be better to do `$j->pretty($prettyr ? 1 : 0)` without `if`.

kwaegema:
Remarks fixed

Issue #255, opened 2017-11-09
CAF::Process supports shell option


stdweird:
currently, `CAF::Process` support the `LC::Process` shell `option` (ie assume there's only one element in the command list, and run it as is (incl lack of protection again shell-isms).
we should eitehr document it or remove it. i'm in favour of removing it, but it goes against the 'should allow to shoot yourself in the foot` principle (also known as `CAF devs shouldn't pretend to know it all` :wink:)

jouvin:
LAL workshop: first step, throw a deprecation warning and at the next workshop, decide to remove the support.

jrha:
N.B. ncm-filecopy appears to use `shell => 1` for restart commands.

jouvin:
Exactly, at least `ncm-filecopy` makes use of it. This seems difficult to remove it, except if we tokenize the command given for the `restart` property in `filecopy`. The goal of the deprecation warning was to better identify all the circumstances where a shell is used. I'm afraid that there is a couple of grid components also rely on this feature but I don't remember precisely.

jrha:
AFAICT it's only `ncm-glitestartup`

jrha:
As discussed we need a migration strategy for (at least) filecopy that does not place a burden on sites when upgrading.

jrha:
No consensus yet, hopefully we can **discuss at workshop**.

jrha:
Discussed at workshop, an option to reduce usage of subshell in filecopy would be to allow a list (or list of lists) of command + arguments for the restart command.

PR #266, opened 2018-07-02
CAF::Process: Deprecate support for executing in a subshell

jrha:
As discussed in #255.


jrha:
Hmpf. JUnit seems to have blown up.

ned21:
In light of [filecopy](https://github.com/quattor/CAF/issues/255#issuecomment-401761636) using this, we probably shouldn't merge it.

ned21:
Given that we've already identified a crucial one, I think this is premature. We don't know how to fix filecopy so it's just going to generate support requests. We need to remediate the components we know about before we start looking for ones we don't!

jouvin:
You are probably right... I think the only possible fix for filecopy is to tokenize the command string. The main advantage over the current situation is probably to avoid running in a shell, avoiding that using a vulnerability somebody can escape to the shell which is running as root. We have to understand if we can do this tokenization in a reasonable way so that it works for almost every command used in our configurations without changing them and that it is easy to fix the other ones....

src/main/perl/Process.pm
```diff
@@ -270,6 +270,10 @@ sub execute
 {
     my $self = shift;
 
+    if ($self->{OPTIONS}->{shell}) {
+        $self->warning('Executing process using subshell, this is deprecated and may be removed in a future release.');
```
stdweird: it's `warn`, not `warning`

stdweird:
maybe instead of `warn`, use verbose (or info), add some `DEVELOPER ONLY FEEDBACK` string to make clear it's not an issue. then after a few months/releases, we can gather usage by grepping logfiles or something similar.
for filecopy, i think we should start deprecating any commands that are not `exe arg1 arg2 ...` (eg with pipe or something nasty) in the schema.

it's clear that for filecopy (or other components that require to run arbitrary code passed via profile) we need a better solution. i'm personnally in favour of replacing such behaviour with a script and run the script (so instead of `command` option in the schema, a `script` option, and write that contents to file and run that file)

jrha:
Ok, bumped to next release, needs discussion. Can we do that in the issue (#255) please?

stdweird:
retest this please

jrha:
No consensus yet, hopefully we can **discuss at workshop**.

PR #273, opened 2020-01-09
Add support for try-restart/condrestart

gombasg:
Unconditionally bringing up a service is not always the correct thing to do.

I've also changed the systemd variant to operate on one unit at a time, see quattor/configuration-modules-core#1376.


jrha (approved):
Looks good AFAICT

stdweird (approved):

PR #275, opened 2022-12-14
add github action for tests

wpoely86:
It fails at this moment: https://github.com/wpoely86/CAF/actions/runs/3592767878

### DIFF
--- a/.github/workflows/continuous-integration.yaml
+++ b/.github/workflows/continuous-integration.yaml
@@ -1,0 +1,44 @@
+name: Run Tests
+
+on: [push, pull_request]
+
+jobs:
+  runtests:
+    runs-on: ubuntu-latest
+    container: rockylinux:8
+    steps:
+    - name: Checkout code
+      uses: actions/checkout@v3
+    - name: Determine hash for caching key
+      id: cachekeystep
+      run: echo "pomcachekey=${{ hashFiles('pom.xml') }}" >> $GITHUB_ENV
+    - name: set up dependencies
+      run: |
+        dnf -y install dnf-plugins-core
+        dnf config-manager --set-enabled appstream
+        dnf config-manager --set-enabled powertools
+        dnf -y install epel-release \
+          http://yum.quattor.org/devel/quattor-yum-repo-2-1.noarch.rpm
+        dnf -y install maven perl-Test-Quattor panc libselinux-utils wget perl-Test-Harness
+    - name: set up template library core from git master
+      run: |
+        cd /tmp
+        # install library core in /tmp, tests need it
+        wget -O template-library-core-master.tar.gz https://codeload.github.com/quattor/template-library-core/tar.gz/master
+        tar -xzf template-library-core-master.tar.gz
+        cd -
+    - name: Cache Maven packages
+      uses: actions/cache@v3
+      with:
+        path: /tmp/m2
+        key: ${{ runner.os }}-m2-${{ env.pomcachekey }}
+        restore-keys: ${{ runner.os }}-m2-
+    - name: run tests
+      run: |
+        mkdir -p /tmp/m2
+        useradd -m -s /bin/bash quattor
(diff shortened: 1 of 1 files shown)


jrha:
Hmm, odd.

src/test/perl/filewriter-notmocked.t
---
This looks like another undefined value problem as seen elsewhere:
```
#   Failed test 'old-style exception thrown'
#   at src/test/perl/filewriter-notmocked.t line 161.
Can't call method "text" on an undefined value at src/test/perl/filewriter-notmocked.t line 162.
```

src/test/perl/path.t
---
These seem to be failing because they can do things they shouldn't be able to.
```
not ok 235 - temp directory on parent without permissions returns undef on failure tempdir
#   Failed test 'temp directory on parent without permissions returns undef on failure tempdir'
#   at src/test/perl/path.t line 657.
```

```
#   Failed test '_reset_exception_fail called 2 times after temp directory creation failure NoAction=0 permission'
#   at /__w/CAF/CAF/src/test/perl/modules/exception_helper.pm line 61.
#          got: '3'
#     expected: '2'
```

```
# VERBOSE Moved target/test/check/move1/src to target/test/check/move2/dst: 1
# DEBUG: 1 Moved src target/test/check/move1/src to dest target/test/check/move2/dst

#   Failed test 'move src target/test/check/move1/src to dest target/test/check/move2/dst w/o backup failed no permission to move src to dest (destdor exists)'
#   at src/test/perl/path.t line 898.
```

```
#   Failed test 'move src target/test/check/move1/src to dest target/test/check/move2/dst with backup '.old' failed no permission to make backup of dest'
#   at src/test/perl/path.t line 910.
```

```
#   Failed test '_reset_exception_fail called 3 times after move failure to cleanup dest with backup'
#   at /__w/CAF/CAF/src/test/perl/modules/exception_helper.pm line 61.
#          got: '1'
#     expected: '3'
```

wpoely86:
> @wpoely86 can you try this one on `rockylinux:8` as well?

Done. Output can be followed in my fork: https://github.com/wpoely86/CAF/actions

jrha (dismissed):
The more I stare at the test output, the more I think that the failures are more to do with how the tests are written. Given that failing tests are better than not running tests, perhaps it would be better to merge this and figure it out as we go.

stdweird:
@wpoely86 can you add in filewriter-notmocked.t before the open on L157
```
diag "EC pre $testdir";
diag `ls -l "$testdir"`, `mount`;
diag explain $EC, " error ", explain $EC->error();
```
and also after the close? (but with `post` instead of pre).

stdweird:
@jrha i suspect some other odd behaviour with filesystem in use (`/__w/`). the file-not-mocked one should really fail

stdweird:
@jrha @wpoely86 the unittests run as root apparently (wwwwhhaaaaattttt ?). anyway, root doesn't care about permission limits so no errors. can you somehow fix this?

.github/workflows/continuous-integration.yaml
```diff
@@ -35,8 +35,9 @@ jobs:
         restore-keys: ${{ runner.os }}-m2-
     - name: run tests
       run: |
-        source /usr/bin/mvn_test.sh
-        mvn_test
+        mkdir -p /tmp/m2
+        chown -R nobody:nobody . /tmp/m2
```
stdweird, line 39: you have no more "regular" user? 
and you shouldn't run any quattor test as root, in any repo.

wpoely86: It's the most regular user present in the container but I can create a fresh one if you like.

I will have a look. I think we run as root in most other tests.

stdweird: nobody is not a regular user, it's normally has not even a login shell

wpoely86:
All seems well now: https://github.com/wpoely86/CAF/actions

jrha (approved):
Amazing, thankyou.

PR #277, opened 2023-08-23
Replace four instances of while-each with foreach

### LINKED ISSUE #109: review the current usage of `each`
`each` is evil, and altough it can be used correctly, should be avoided at all costs. after learning the lesson once more (i ran into it previously, but somehow forgot), we should get rid of it.

the problem is that each sort-of remembers the last element it iterated over.
this has all sorts of nasty side effects, but the following example illustrates the danger of using each and stopping the iteration.

``` perl
#!/usr/bin/perl

use strict;
use warnings;

my %h = (
    a => 1,
    b => 2,
    );

sub bad
{
    my $stop = shift;
    while (my ($k, $v) = each(%h)) {
        print "$k $v\n";
        return if $stop;
    }
}

sub good
{
    my $stop = shift;
    foreach my $k (keys%h) {
        print "$k $h{$k}\n";
        return if $stop;
    }
}


print "bad\n";
bad(1);
bad();

print "good\n";
good(1);
good();
```

which gives output

```
bad
b 2
a 1
good
b 2
b 2
a 1
```

jrha:
AFAICT this preserves existing behaviour.

Resolves #109.

### DIFF
--- a/src/main/perl/ReporterMany.pm
+++ b/src/main/perl/ReporterMany.pm
@@ -30,8 +30,8 @@ sub _rep_setup
     # initialise with current "global" CAF::Reporter settings
     #   if DEBUGLV is not set
     if (! exists($self->{$DEBUGLV})) {
-        while (my ($opt, $val) = each (%$CAF::Reporter::_REP_SETUP)) {
-            $self->{$opt} = $val;
+        foreach my $opt (keys %{$CAF::Reporter::_REP_SETUP}) {
+            $self->{$opt} = ${$CAF::Reporter::_REP_SETUP}{$opt};
         }
     }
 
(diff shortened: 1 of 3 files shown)


ned21:
hmm.. tricky syntax!

I think it would be good if the reason for the removal was documented in the commit, otherwise I have to find the PR that merged it, then reference the issue. You could just summarise it as:

```
Replace while-each with foreach

`each` sort-of remembers the last element it iterated over which can lead to unexpected results. 
While it can used correctly, it was decided to amend our coding style use prefer foreach.
Fixes: #109 
```

I think this could also be one commit rather than 3, although 3 is easier to revert if we find a problem with one of changes.

ned21 (approved):

PR #279, opened 2024-10-29
ServiceActions: Unescape daemon names

jrha:
This allows for use with templated services e.g. `ssh@mgmt.service` as `@` is an invalid character in a dict key.

Fixes quattor/configuration-modules-core#1672.


stdweird:
can you add a unittest (or adapt one)?

stdweird (approved):

